PR #3, opened 2022-06-09
Image info tests

lavocatt:
This PR brings up a new set of tool:
* `image-info` is a tool to generate a specific JSON from an image built by `osbuild`
* `import-image-test` is a tool used to import the manifest data base from `composer` and generate test cases where an input manifest and a resulting image-info reside in the same JSON.
* `osbuild-image-test` is a tool that takes a JSON produced by `import-image-test`, then uses `osbuild` for input's manifest, then uses `image-info` to generate the `image-info` from the resulting image and finally compares it from the `image-info` from the input JSON. This tool takes a directory in input as well as some filter to run specific archs and distros.

Then this PR introduces a new CI test that runs `osbuild-image-test` alongside what is found under ` manifest-db`. It will error out if the produced `image-info` differs from what was in the db. It will produce as artifacts a set of new `image-info` that can later be used either to compare what is going on if there are differences, or to update the db with fresh builds.


lavocatt:
What needs to be done soon:

- [x] Run the CI against all archs/distros 
- [x] export the resulting image-info as a CI artifact -> Done
- [x] schutzbot status too ?

gicmo:
I think I would move the `selinux-context.sh` script into `schutzbot`, it is not really a tool but more of a env fix.

PR #5, opened 2022-06-17
Image info tests

lavocatt:
This PR brings up a new set of tool:
* `image-info` is a tool to generate a specific JSON from an image built by `osbuild`
* `import-image-test` is a tool used to import the manifest data base from `composer` and generate test cases where an input manifest and a resulting image-info reside in the same JSON.
* `osbuild-image-test` is a tool that takes a JSON produced by `import-image-test`, then uses `osbuild` for input's manifest, then uses `image-info` to generate the `image-info` from the resulting image and finally compares it from the `image-info` from the input JSON. This tool takes a directory in input as well as some filter to run specific archs and distros.

Then this PR introduces a new CI test that runs `osbuild-image-test` alongside what is found under ` manifest-db`. It will error out if the produced `image-info` differs from what was in the db. It will produce as artifacts a set of new `image-info` that can later be used either to compare what is going on if there are differences, or to update the db with fresh builds.


schutzbot/team_ssh_keys.txt
```diff
@@ -0,0 +1,10 @@
...
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDMk4+D+GkL/se9RYdQ6mmv0v7cG3gXPiZdLvsbRFk+JMgnLEFUu4uwRN9+huMWdTXYLVcCZVw1l9kmDAOwPKYCGwetOwYI73nYt4n7i9zSQqKknxEO2w+16PvWy6kbgk7zQp9YEqJlIeZbVXiDuFeFyqEH5TxdAfDLu0VG/Db0M1Cor3v7g0/utHWVfS0QiLMQ5xLZ3jAbn/uTjOzsvtA219BTfdx7IVxKmG9/zeLUqyHBcOquI2rUjgHj1FN4yMCob2pSmAfV2u1iQZYHghXlw/kzh4BODjOSWCO9v+621eBdDJ36hdMHWizevxLz9Tj7X1Pwz+4TS/WABXs5MXBFcxx65wpSBr84LfZfWebe+vFe3Kkmdop3FPt7uuZyl8xdRsvfoIG3qZg2izX4IAtXo9tR++yIsEkP/5k/NVwkq158molT0HUruuK9regPRRvf7j6psAH4bMhV3o430eD7ibC5g1yT3kKbGnbJJX+HHL0+lt8IKMPCHWEWOT8uR0x9VUbT4RYFXt53I/ZXHXop+eypEctoKlsgMv1rpPGwSnrhtK1hgf399ph7GDjixZiuzcTP+Ulmj3atB32UHR7mRjarnaCI1w7PJt8pmg9K+8/XUa51q7bap1PGwsJS0lZ3zJEXFPdJfqO8/fkhDb4NYPlX0/bg/LNI9YqPg8CWSw== jkozol-2@redhat.com
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDY/ylCrPBzil4TnZR4tWULpz3QgfBMQyEnMOHDAJNp/FK70hD+PUiRm3UY96pmGXonQvqiDoyPuVh025FkWshPK91Dyq8QD8h25q5C5Cg6kMgBpdGzbX44ksms1KyOHmSZ48MpWw3PFOrlNP1vysr6Imjz9Jixmx4sOZvqKnrbsbOW04gowVzpZM8m048lvf6/KhqeImfeSRc9Rtpos8GqEQVlwRevE1qBON963V1QtFOrm9weoQgb369SdqRRdxaGNAymNh3d78DneOWXmEyBflLSpIDx5I2s/1NB1Dp95Bp3VvlV3CH1HC7LAFKYi+xsz3/KHdgtvgShX6LFSdsp rvykydal@dhcp-lab-144.englab.brq.redhat.com
```
achilleas-k: I guess these are from osbuild?  Maybe we should combine the osbulid-composer ones in here too.
Perhaps we can have a central place for it, but that's not relevant to this PR.

tools/osbuild-image-test
```diff
@@ -413,6 +418,12 @@ def main():
         type=os.path.abspath,
         default="/var/lib/osbuild/out",
         help="directory where result objects are stored")
+    parser.add_argument(
+        "--imi-output",
```
achilleas-k: `imi`?  Is that short for `image-info`?
It's a big unclear.  I know `--image-info-output` is a bit long, but at least it's clear.

tools/osbuild-image-test
```diff
@@ -413,6 +418,12 @@ def main():
...
+        metavar="DIRECTORY",
+        type=os.path.abspath,
+        default="imis",
```
achilleas-k: Similar comment as above.  `imis` as a directory name isn't really saving us anything.  Better to be clear.

achilleas-k (commented):
Some very minor in-line comments.
I'm good to merge this if those are done, if nothing else the `imi` command line flag and directory name.

In general though, I think we (@gicmo, @thozza, and I) feel that this repo shouldn't need to do any testing in itself.  The scripts, manifests, and image info we want, so we can merge as is and then work on the what the CI does in later iterations.

Here's how we envision it:
- We want to import manifests from `osbuild-composer` into this repo on a schedule :heavy_check_mark: 
- We want to generate image-info from those manifests using a specific (pinned) version of `osbuild` with the version of `image-info` that is in this repository :heavy_check_mark: 
- We feel that this repo should **not** have tests that fail on changes.  PRs that come into this repository **should** be for changing manifests and image info.  This repository / db should be a collection of manifests that relate to a given version of `osbuild-composer` and a collection of image-info that relate to a given version of manifests + `osbuild`.
-  **IF** (for example) we have a situation where a change in `osbuild` produces a change in `image-info` even though nothing else changed, then that should fail in `osbuild`'s CI, not here.

The tooling and scaffolding to make this happen is mostly done in this PR, so we can merge it for now and work on it.  Perhaps the most important thing we need immediately after is for the bot/CI to create PRs on this repo when things change.
A common workflow might be:
- Trigger a GitHub action manually (like the rpmrepo snapshots)
- The CI runner pulls in the manifests from `osbuild-composer` (or creates them from scratch on `main` using `gen-manifests`)
- Builds every image
- Produces the image info for all images
- Opens a PR on this repository to update all the information.
- We review and merge.

schutzbot/team_ssh_keys.txt
lavocatt: I starting to think (also not relevant to this PR) that the schutzbot/ folder should be a repo in itself. And that we should import it as a submodule. We start to have a few projects running a gitlab ci now, and each one adds up some complexity to this thing, and that might go wrong at some point.

achilleas-k (approved):
All good.  Let's merge and iterate in follow-up PRs.

PR #7, opened 2022-06-27
tools/osbuild-image-test: fixes

lavocatt:
Fix a few errors left in the previous PR due to lack of testing.

### DIFF
--- a/tools/osbuild-image-test
+++ b/tools/osbuild-image-test
@@ -471,7 +471,7 @@ def main():
         "--skip-inspection",
         action="store_true",
         default=False,
-        help="Disengage comparison of  the built images to their reference in the DB"
+        help="Disengage comparison of the built images to their reference in the DB"
     )
 
     args = parser.parse_args()
--- a/tools/osbuild-image-test
+++ b/tools/osbuild-image-test
@@ -27,41 +27,26 @@ RED = "\033[31m"
 YELLOW = "\033[33m"
 
 
-def show_validation(result, name="validation"):
-
-    print(f"{BOLD}{name}{RESET} ", end='')
-
-    if result:
-        print(f"is {BOLD}{GREEN}valid{RESET}")
-        return
-
-    print(f"has {BOLD}{RED}errors{RESET}:")
-    print("")
-
-    for error in result:
-        print(f"{BOLD}{error.id}{RESET}:")
-        print(f"  {error.message}\n")
-
-
 class OSBuild:
-    def __init__(self, path, store, outdir):
+    """
+    Takes care of building images using OSBuild
+    """
+
+    def __init__(self, path, store, outdir, verbose):
         self.path = path
         self.store = store
         self.outdir = outdir
         self.checkpoints = []
-        self.verbosity = 1
+        self.verbosity = 2 if verbose else 1
 
     def compile(self, manifest: dict, exports: List, checkpoints=None):
-        res = self.run([],
+        res = self._run([],
                        manifest=manifest,
                        exports=exports,
                        checkpoints=checkpoints)
         return res
 
-    def inspect(self, manifest: dict) -> dict:
-        return self.run(args=["--inspect"], manifest=manifest)
-
-    def run(self,
+    def _run(self,
             args: Optional[List[str]] = None,
             manifest: Optional[dict] = None,
             exports: Optional[List] = None,
@@ -122,9 +107,13 @@ class OSBuild:
 
 
 class ImageInfo:
-    def __init__(self, path: str):
+    """
+    Invokes the image-info executable
+    """
+
+    def __init__(self, path: str, verbose):
         self.path = path
-        self.verbosity = 1
+        self.verbosity = 2 if verbose else 1
 
     def run(self, target: str):
         cmd = [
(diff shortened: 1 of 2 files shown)


tools/osbuild-image-test
```diff
@@ -469,7 +470,7 @@ def main():
     parser.add_argument(
         "--skip-inspection",
         action="store_true",
-        default=True,
+        default=False,
         help="Disengage comparison of  the built images to their reference in the DB"
```
achilleas-k: Noticed a double space here between `of` and `the`

tools/osbuild-image-test
```diff
@@ -300,13 +300,14 @@ def inspect_image(test: TestCase, imi: ImageInfo, args: argparse.Namespace) -> b
 
     write_json(test.id, result, args.imi_output)
 
-    if not test.image_info:
-        print(f" {YELLOW}Warning{RESET}: No image info to compare to; ", end="")
-        return True
+    if not args.skip_inspection :
```
achilleas-k: Bad codestyle: Space between name and colon.

tools/osbuild-image-test
```diff
@@ -300,13 +300,14 @@ def inspect_image(test: TestCase, imi: ImageInfo, args: argparse.Namespace) -> b
...
+        if not test.image_info:
+            print(f" {YELLOW}Warning{RESET}: No image info to compare to; ", end="")
+            return True
 
-    if result != test.image_info:
-        test.error = "image-info mismatch"
-        return False
+        if result != test.image_info:
+            test.error = "image-info mismatch"
+            return False
```
achilleas-k: If I'm reading this correctly, the `skip-inspection` flag only skips the comparison with the old image info, but the script will still read the old info, it just wont do anything with it?
Seems unnecessary and also hard to read.

This could use a rewrite to make things clearer.  If we don't need the old results and this flag "turns the tool
into an image-info generator", then it should just be that and not assume (or care) about any old image info.

achilleas-k (changes requested):

tools/osbuild-image-test
lavocatt: Yes, this is quite bad indeed. I need to rewrite this.

tools/osbuild-image-test
```diff
@@ -501,10 +478,44 @@ def main():
...
+            # with what's in the DB.
+            if not t.image_info:
+                print(f" {YELLOW}Warning (ignoring){RESET}: No image info to compare to")
+                continue
+
+        if build_test(t, osb, imi, args):
+            # Every generated image-info needs to be written to disk.
+            write_json(t.id, t.result, args.image_info_output)
+            if not args.generator_mode:
+                inspect_image(t, imi, args)
```
achilleas-k: The output is never used.  This is a no-op.

achilleas-k (changes requested):
Much cleaner.  Bit of an issue below.

tools/osbuild-image-test
lavocatt: The output is used. inspect_image writes down its result to the test object, and if an error occurs, test.error is set and tested afterward.

lavocatt:
I think were good to merge this:

https://gitlab.com/redhat/services/products/image-builder/ci/osbuild/-/jobs/2652246943 <- in this job, on osbuild, we have the desired output in the console:
```
centos_8-aarch64-ami-boot Building & Comparing: SUCCESS
centos_8-aarch64-edge_commit-boot Building & Comparing: SUCCESS
```

meaning it builds and compare so, that's good
And for manifest-db https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/jobs/2650992806 <- in this job, we also have the desired output:
```
centos_8-x86_64-ami-boot Building only: SUCCESS
centos_8-x86_64-edge_commit-boot Building only: SUCCESS
```

tools/osbuild-image-test
```diff
@@ -501,10 +456,49 @@ def main():
...
+            print("Building only:", end=" ", flush=True)
+        else:
+            print("Building & Comparing:", end=" ", flush=True)
+            # If the program is configured to build & compare, then if the
+            # test case haven't an image-info to compare to, it's useless to
+            # build it. Because in this mode we are only interested in comparing
+            # with what's in the DB.
+            if not t.image_info:
+                print(f" {YELLOW}Warning (ignoring){RESET}: No image info to compare to")
+                continue
```
achilleas-k, line 504: If the script is called to compare and there's nothing to compare to, I think that should be an error.  Since generator mode exists as an explicit option, I falling back to just generating when the task is to compare seems like a good way to miss errors.

tools/osbuild-image-test
```diff
@@ -204,132 +221,143 @@ class TestCase:
...
 
-    # for now we just assume the last pipeline is the export
-    return [next(reversed(pipelines))]
+        build = manifest.get("build")
+        if build:
+            checkpoints += [build.name]
 
+        commit = manifest.get("ostree-commit")
+        if commit:
+            checkpoints += [commit.name]
```
achilleas-k, line 241: We could checkpoint everything, which would be great for situations where manifests change in late stages only.

Future improvement though.

achilleas-k (approved):
Couple of comments but I'd be okay to merge and fix them up later if you prefer.

tools/osbuild-image-test
lavocatt: The idea now, is to let `osbuild` succeed against a non complete `DB`. Maybe in a follow up we can make a parameter `--allow-empty`.

PR #9, opened 2022-07-01
mypy: include mypy testing for the repository

lavocatt:
Run mypy for all files under tools/ and test/cases/
Fix the type errors in the files.

### DIFF
--- a/tools/osbuild-image-test
+++ b/tools/osbuild-image-test
@@ -14,7 +14,7 @@ import traceback
 import shutil
 
 from collections import defaultdict
-from typing import Iterator, List, Optional
+from typing import Iterator, List, Optional, Dict
 
 import osbuild
 import osbuild.meta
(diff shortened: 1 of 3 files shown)


achilleas-k (approved):

tools/osbuild-image-test
```diff
@@ -41,16 +41,16 @@ class OSBuild:
 
     def compile(self, manifest: dict, exports: List, checkpoints=None):
```
supakeen: Using the lowercase `dict` here is a 3.9+ only feature and likely doesn't do what we want < 3.9.

supakeen (changes requested):
This all looks good. Could you get around the # type-ignore by passing --ignore-missing-imports to mypy or is that not a thing? The only thing I saw is the lowercase dict.

tools/osbuild-image-test
lavocatt: Done :) Thanks!

lavocatt:
> This all looks good. Could you get around the # type-ignore by passing --ignore-missing-imports to mypy or is that not a thing? The only thing I saw is the lowercase dict.

Indeed that was better with this option.

tools/osbuild-image-test
```diff
@@ -321,9 +326,9 @@ class TestCases:
 
     @staticmethod
     def _print_stats(tests: List[TestCase]) -> None:
-        distros = defaultdict(int)
-        archs = defaultdict(int)
-        types = defaultdict(int)
+        distros: Dict[str, Any] = defaultdict(int)
+        archs: Dict[str, Any] = defaultdict(int)
+        types: Dict[str, Any] = defaultdict(int)
```
supakeen: Wouldn't the `defaultdict(int)` here indicate that it should be `Dict[str, int]`? :)

tools/osbuild-image-test
```diff
@@ -39,18 +39,18 @@ class OSBuild:
...
+                        checkpoints=checkpoints)
         return res
 
     def _run(self,
-            args: Optional[List[str]] = None,
-            manifest: Optional[dict] = None,
-            exports: Optional[List] = None,
-            checkpoints: Optional[List] = None):
+             args: Optional[List[str]] = None,
+             manifest: Optional[dict] = None,
```
supakeen, line 51: One last lowercased `dict`.

supakeen (commented):
Two tiny things I spotted.

supakeen (approved):

tools/osbuild-image-test
lavocatt: :D thanks!

PR #10, opened 2022-07-04
Regression testing

lavocatt:
add a regression test to the suite
    
In this test the tool osbuild-image-test run against two databases. The first one containing a successible test case and the  second one a faulty test cases that will fail upon comparison of the image-info.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -35,6 +35,22 @@ init:
     - schutzbot/update_github_status.sh start
   interruptible: true
 
+Regression:
+  stage: test
+  extends: .terraform
+  script:
+    - schutzbot/deploy.sh
+    - sudo schutzbot/selinux-context.sh
+    - sudo test/cases/regression_tests
+  artifacts:
+    when: always
+    paths:
+      - generated-image-infos/
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/fedora-35-x86_64
+
 Manifests:
   stage: test
   extends: .terraform
--- a/test/cases/regression_tests
+++ b/test/cases/regression_tests
@@ -1,0 +1,51 @@
+#!/usr/bin/python3
+"""
+Tests the osbuild-image-test by running it against two fake DBs. This test must
+run on Fedora35 x86_64
+"""
+
+import subprocess
+import configparser
+import platform
+import sys
+from subprocess import CalledProcessError
+
+OS_RELEASE_PATH = "/etc/os-release"
+
+config = configparser.ConfigParser()
+with open(OS_RELEASE_PATH, 'r', encoding='utf-8') as f:
+    config_string = '[DEFAULT]\n' + f.read()
+config.read_string(config_string)
+distro = f"{config.get('DEFAULT', 'ID')}-{config.get('DEFAULT', 'VERSION_ID')}"
+distro = distro.replace('"', '')
+distro = distro.replace('.', '')
+
+print(f"Testing the osbuild-image-test for arch {platform.machine()} and "
+        f"distribution {distro} on folder test/data/db-success "
+        "this test must succeed", flush=True)
+
+try:
+    subprocess.run(["tools/osbuild-image-test",
+                    f"--arch={platform.machine()}",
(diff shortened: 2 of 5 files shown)


test/cases/regression_tests
```diff
@@ -0,0 +1,51 @@
...
+try:
+    subprocess.run(["tools/osbuild-image-test",
+                    f"--arch={platform.machine()}",
+                    f"--distro={distro}",
+                    "--db-path=test/data/db-success"],
+                    check=True)
+except CalledProcessError:
+    sys.exit(1)
+else:
+    print("SUCCESS")
```
achilleas-k, line 51: Where do we check for the expected failure?

achilleas-k: Should this be run on `db-failure` and the condition flipped?

lavocatt: Ahah! Thanks, yes clearly that's a half copy/paste issue here.

test/cases/regression_tests
```diff
@@ -0,0 +1,51 @@
...
+try:
+    subprocess.run(["tools/osbuild-image-test",
+                    f"--arch={platform.machine()}",
+                    f"--distro={distro}",
+                    "--db-path=test/data/db-failure"],
+                    check=True)
+except CalledProcessError:
+    print("SUCCESS")
+else:
+    sys.exit(1)
```
achilleas-k, line 51: This is probably fine for now but it might be good to have a specific error we can catch to make sure there wasn't some other kind of issue.  Like if the `db-failure` goes away or gets renamed or something else happens that might cause an error in the script, we're be oblivious to it because we're just expecting the script invocation to fail with a non-zero exit code.

lavocatt: Yes, I wanted to have proper unit tests at the beginning, but since `osbuild-image-test` is not a python module, you can't really import it to run smarter tests. I could try to setup different return values for the executable and test that I get the right one. Do you have other ideas ?

achilleas-k: Doesn't need to be too sophisticated I think.  If the test script prints error messages we could just check for the expected one.

lavocatt: I'll try to figure out something else better in a follow up then :)

.gitlab-ci.yml
```diff
@@ -35,6 +35,22 @@ init:
...
+    - sudo schutzbot/selinux-context.sh
+    - sudo test/cases/regression_tests
+  artifacts:
+    when: always
+    paths:
+      - generated-image-infos/
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/fedora-35-x86_64
```
thozza, line 52: Wouldn't it make more sense to use Fedora 36 instead? F35 will be EOL in November 2022, so it would save us some work for longer time.

thozza (commented):
All looks good, but I think that you should default to the latest Fedora release for the test.

.gitlab-ci.yml
lavocatt: Not yet unfortunately. I need to land the updating proceedure that depends on this one. Because there's no image-info compiled for fedora-36 in the DB.
So first I need to land this PR, then the PR to update the DB. Then activate runners for fedora36, then update again the DB, then update the regression tests :)

lavocatt: I'll open an issue here to remember to do that.

thozza: What's the benefit of landing this now in that case? 🤔 Seems that it just creates more work for you.

lavocatt: So, I want to be sure that next commits don't break anything. So I want at least to have one success build and one failed build for all the PR after this one. With the upcoming of automated updates, the image info generation is removed from the CI tests and only exist inside the triggered special pipelines to build them upon update of the manifests, which means that, there would be no test at all to ensure that nothing is broken upon a classic PR.

lavocatt: So I kind of "need" that.

thozza (approved):
While I admit that I still don't fully understand your reasons to get this merged it now, before the other work, I don't see any point in blocking you on this.

👍

PR #16, opened 2022-07-05
Image info update

lavocatt:
Add a framework to update the image-info in the repository.
This PR brings several tools to do the task:

* The first one is `ci_import` is used to import genereated image-info from a pipeline job in the CI
* The second one is a github action that can be trigger to import the manifests from the last composer version, this action is manually triggered and will create a PR containing a single commit hosting the changes to the manifests.
* The third and last one is a github action that is triggered for each new PR. It will call the `ci_import` tool after `Schutzbot` has successfully run. If new image-info are produced, or if they are updated, then a new commit will land on top of PR with these modifications.

Due to the properties of the gthub actions, the manual trigger could not be tested here, however, it runs on this ci job:
https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/610538989

### DIFF
--- a/.github/workflows/auto_update.yml
+++ b/.github/workflows/auto_update.yml
@@ -1,0 +1,49 @@
+name: Update DB
+
+on: workflow_dispatch
+
+jobs:
+  update_db:
+    name: "Update DB"
+    runs-on: ubuntu-latest
+    env:
+      IMAGEBUILDER_BOT_GITLAB_SSH_KEY: ${{ secrets.IMAGEBUILDER_BOT_GITLAB_SSH_KEY }}
+      TRIGGER_TOKEN: ${{ secrets.TRIGGER_TOKEN }}
+    steps:
+      - name: Clone composer
+        uses: actions/checkout@v3
+        with:
+          repository: osbuild/osbuild-composer
+          path: composer
+
+      - name: Checkout manifest-db
+        uses: actions/checkout@v3
+        with:
+          path: manifest-db
+          fetch-depth: 0
+
+      - name: Update DB
+        run: |
+          ./manifest-db/tools/import-image-tests composer/test/data/manifests/ manifest-db/manifest-db --manifest-only --db-ignore=manifest-db/db-ignore --verbose
+
+      - name: Send to CI
+        run: |
+          cd manifest-db
+          git config --local user.name "SchutzBot"
+          git config --local user.email "imagebuilder-bots+schutzbot@redhat.com"
+
+          now=$(date '+%Y-%m-%d-%H%M%S')
+          BRANCH_NAME="db-update-$now"
+          git checkout -b "$BRANCH_NAME"
+          git add -A
+          git commit -m "db: automatic update of manifests."
+
+          mkdir -p ~/.ssh
+          echo "${IMAGEBUILDER_BOT_GITLAB_SSH_KEY}" > ~/.ssh/id_rsa
+          chmod 400 ~/.ssh/id_rsa
+          touch ~/.ssh/known_hosts
+          ssh-keyscan -t rsa gitlab.com >> ~/.ssh/known_hosts
+          git remote add ci git@gitlab.com:redhat/services/products/image-builder/ci/manifest-db.git
+          git push -f ci -o ci.skip
+
+          curl --request POST --form token=$TRIGGER_TOKEN --form ref=$BRANCH_NAME https://gitlab.com/api/v4/projects/36844106/trigger/pipeline
--- a/README.md
+++ b/README.md
@@ -1,0 +1,29 @@
+Manifest-db
+=======
+
+Stores all the manifests and corresponding image-infos to test OSbuild.
+
+### Update workflow
+
+The update workflow starts with a github action that is manually triggered. This
+GH action is named `Update DB`. It will firstly clone the latest
+`osbuild-composer` and copy all its manifests into the database. Secondly the
+action will create a commit and push this on the corresponding gitlab ci
+project
+(https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db).
+Thirdly, after the push is completed, the action triggers a special pipeline on
+the CI. This pipeline builds every manifest from the DB. Once building is done,
+the pipeline starts the `schutzbot/include_image_info.sh` script. This script
+will download all the produced image-infos resulting from the builds and update
+the DB with them. Then it will amend the commit sent to the CI pipeline with
+these freshly built image-info, push this commit to github on a branch and
+create a pull request out of it.
+
+### Contributing
+
+Please refer to the [developer guide](https://www.osbuild.org/guides/developer-guide/developer-guide.html) to learn about our workflow, code style and more.
+
+### License:
+
+ - **Apache-2.0**
+ - See LICENSE file for details.
--- a/db-ignore
+++ b/db-ignore
@@ -1,0 +1,10 @@
+centos_8-x86_64-image_installer-boot
+centos_8-x86_64-image_installer_with_users-boot
+centos_9-x86_64-image_installer-boot
+centos_9-x86_64-image_installer_with_users-boot
+rhel_85-x86_64-image_installer-boot
+rhel_85-x86_64-image_installer_with_users-boot
+rhel_86-x86_64-image_installer-boot
+rhel_86-x86_64-image_installer_with_users-boot
+rhel_90-x86_64-image_installer-boot
+rhel_90-x86_64-image_installer_with_users-boot
--- a/schutzbot/include_image_info.sh
+++ b/schutzbot/include_image_info.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+set -euxo pipefail
+sudo dnf install gh -y pip python3-gitlab
+
+# Login to GH
+echo "${SCHUTZBOT_GH_TOKEN}" | gh auth login --with-token
+
+# import the image info from the current build
+./tools/ci_import --pipeline-id "$CI_PIPELINE_ID" --token "${GITLAB_TOKEN}"
+
+#remove ci-details-before-run
+mv ci-details-before-run /tmp
+rm -rf generated-image-infos
+
+# Amend the commit containing the db update
+git checkout "$CI_COMMIT_BRANCH"
+
+git config --local user.name "SchutzBot"
+git config --local user.email "imagebuilder-bots+schutzbot@redhat.com"
+
+git remote add upstream https://schutzbot:"$SCHUTZBOT_GH_TOKEN"@github.com/schutzbot/manifest-db.git
+
+git add -A && \
+    git commit --amend -m "db: update
+
+Automatic update:
+- manifests from latest composer
+- image-info from pipeline $CI_PIPELINE_ID" && \
+    git push --set-upstream upstream "$CI_COMMIT_BRANCH" && \
+    gh pr create \
+        --title "db update" \
+        --body "automated db update" \
+        --repo "osbuild/manifest-db" \
+        -r lavocatt
+
+#restore ci-details-before-run
+mv /tmp/ci-details-before-run .
--- a/tools/ci_import
+++ b/tools/ci_import
@@ -1,0 +1,122 @@
+#!/bin/env python
+
+import argparse
+import json
+import os
+import sys
+import zipfile
+import tempfile
+from os import path
+
+import gitlab
+
+IMI_DIRECTORY = "generated-image-infos"
+MANIFEST_DB_PROJECT_ID = 36844106
+
+
+def list_jobs(pipeline, token):
+    """
+    Connects to the gitlab API and retrives the list of the jobs contained in a
+    given pipeline.
+    """
+    gl = gitlab.Gitlab("https://gitlab.com", oauth_token=token)
+    project = gl.projects.get(MANIFEST_DB_PROJECT_ID)
+    pipeline = project.pipelines.get(pipeline)
+    return pipeline.jobs.list(all=True), project
+
+
+def get_job_dir(tempdir, job):
+    """
+    Returns a directory where to store a job's artifacts
+    """
+    directory = path.join(tempdir, str(job.get_id()))
+    os.makedirs(directory, exist_ok=True)
+    return directory
+
+
+def download_artifacts(directory, project, job):
+    """
+    Download an artificact.zip if available for a given job and extract it.
+    Can raise an exception if no artifacts are available.
+    """
+    filename = os.path.join(directory, "artifact.zip")
+    with open(filename, "wb") as f:
+        project.jobs.get(job.get_id()).artifacts(streamed=True, action=f.write)
+    z = zipfile.ZipFile(filename)
+    z.extractall(directory)
+    print(f"    downloading artifact.zip to {filename} for job {job.get_id()}")
+
+
+def download_image_infos(job_dir, project, job):
+    """
+    Downloads the image info and return their path as a list
+    """
+    try:
+        download_artifacts(job_dir, project, job)
+    except gitlab.exceptions.GitlabGetError as e:
+        print(f"    {e} ⚠️")
+        return []
+    image_info_directory = os.path.join(job_dir, IMI_DIRECTORY)
+    try:
+        image_infos = os.listdir(image_info_directory)
+        print(f"    image information imported for {job.get_id()}")
+        return [os.path.join(image_info_directory, image_info) for image_info in image_infos]
+    except FileNotFoundError:
+        print(f"    no image information for {job.get_id()}")
+        return []
+
+
+def update_db(image_info_path):
+    """
+    Take a downloaded image info and write its content into the corresponding
+    test case in the database.
+    """
+    image_info = os.path.basename(image_info_path)
+    print(f"    {image_info}:", end=" ")
+    # load the image info content
+    with open(image_info_path, "r", encoding="utf-8") as f:
+        image_info_data = json.load(f)
+    # load test case
+    test_case_path = os.path.join("manifest-db", image_info)
+    with open(test_case_path, "r", encoding="utf-8") as f:
+        tc_data = json.load(f)
+    # update test case
+    print(f"updating {test_case_path}:", end=" ")
+    tc_data["image-info"] = image_info_data
+    with open(test_case_path, "w", encoding="utf-8") as f:
+        json.dump(tc_data, f, indent=2)
+    print("✅")
+
+
+def import_image_infos(tempdir, pipeline, token):
+    """
+    Imports the image info from a pipeline into the database.
+    """
+    jobs, project = list_jobs(pipeline, token)
+    for job in jobs:
+        print(f"Job {job.get_id()}:")
+        # for each downloaded image info, update the test cases
+        for image_info_path in download_image_infos(get_job_dir(tempdir, job), project, job):
+            update_db(image_info_path)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Import image-info from ci jobs")
+    parser.add_argument(
+        "--pipeline-id",
+        required=True,
+        help="Id of the pipeline on gitlab"
+    )
+    parser.add_argument(
+        "--token",
+        required=True,
+        help="gitlab token"
+    )
+
+    args = parser.parse_args()
+    with tempfile.TemporaryDirectory() as tempdir:
+        import_image_infos(tempdir, args.pipeline_id, args.token)
+
+
+if __name__ == "__main__":
+    sys.exit(main())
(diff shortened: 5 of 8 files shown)


tools/ci_import
```diff
@@ -0,0 +1,94 @@
...
+
+import argparse
+import json
+import os
+import sys
+import zipfile
+
+import gitlab
+
+IMI_FOLDER = "generated-image-infos"
```
ondrejbudai: nitpick: `directory` is more idiomatic in linux.

tools/ci_import
```diff
@@ -0,0 +1,94 @@
...
+
+def import_image_infos(pipeline, token, verbose=False):
+    gl = gitlab.Gitlab("https://gitlab.com", oauth_token=token)
+    project = gl.projects.get(MANIFEST_DB_PROJECT_ID)
+    pipeline = project.pipelines.get(pipeline)
+    jobs = pipeline.jobs.list(all=True)
+    for job in jobs:
+        if verbose:
+            print(f"Job {job.get_id()}:")
+        folder = f"/tmp/{job.get_id()}/"
```
ondrejbudai: `path.join` feels cleaner. Might be even worth switching to `pathlib` in this script, as it does a fair amount of path manipulation.

(also: `folder` once again)

ondrejbudai: One more thing: I think we should do this in `tempfile.TemporaryDirectory` to prevent any conflicts.

tools/ci_import
```diff
@@ -0,0 +1,94 @@
...
+                tc_data["image-info"] = imi_data
+                try:
+                    if verbose:
+                        print(f"        updating {test_case_path}:", end=" ")
+                    with open(test_case_path, "w", encoding="utf-8") as f:
+                        json.dump(tc_data, f, indent=2)
+                    if verbose:
+                        print("✅")
+                except Exception as e:  # pylint: disable=[broad-except]
+                    if verbose:
```
ondrejbudai: Why is this exception ignored? I think that this script is unsuccessful if something fails in this `try` block.

tools/ci_import
```diff
@@ -0,0 +1,94 @@
...
+                        print(f"{e} ❌")
+        except gitlab.exceptions.GitlabGetError as e:
+            if verbose:
```
ondrejbudai: Same concern about an ignored exception here.

tools/import-image-tests
```diff
@@ -63,9 +63,16 @@ def load_test(name, fullpath) -> dict:
     return test
 
 
-def export_test(test, target):
+def export_test(test, target, manifest_only):
     name = test["id"]
     fullpath = os.path.join(target, name + ".json")
+    if  manifest_only and os.path.exists(fullpath):
```
ondrejbudai: Double-space after `if`. We should definitely invest in a linter for this project in the near future. :-)

tools/import-image-tests
```diff
@@ -135,13 +135,28 @@ def main():
         const=0,
         dest="verbosity",
         help="suppress output")
+    parser.add_argument(
+        "--db-ignore",
+        type=os.path.abspath,
+        default=None,
+        help="A file containing one per ligne files name to not import")
```
ondrejbudai: I'm not 100% sure, but this sentence should be written like this:

`A file containing one per line file names not to import`

tools/import-image-tests
```diff
@@ -135,13 +135,28 @@ def main():
...
 
     args = parser.parse_args()
 
+    db_ignore = []
+    if args.db_ignore:
+        with open(args.db_ignore, "r", encoding="utf-8") as f:
+            db_ignore = f.readlines()
```
ondrejbudai: What about using a simpler `f.read().splitlines()` and omit the next line?

.github/workflows/auto_update.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Checkout manifest-db
+        uses: actions/checkout@v3
+        with:
+          path: manifest-db
+          fetch-depth: 0
+
+      - name: Update DB
+        run: |
+          ./manifest-db/tools/import-image-tests --verbose composer/test/data/manifests/ manifest-db/manifest-db --manifest-only --db-ignore=manifest-db/db-ignore
```
ondrejbudai: I suggest moving the positional arguments (`composer/test/data/manifests/` and `manifest-db/manifest-db`). Currently, it looks like that `composer/test/data/manifests/` is a value of `--verbose`)

.github/workflows/auto_update.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Send to CI
+        run: |
+          cd manifest-db
+          git config --local user.name "SchutzBot"
+          git config --local user.email "schutzbot@redhat.com"
```
ondrejbudai: Sadly, we weren't able to get this address, please use `imagebuilder-bots+schutzbot@redhat.com`

.github/workflows/auto_update.yml
```diff
@@ -0,0 +1,52 @@
...
+
+          now=$(date '+%Y-%m-%d-%H%M%S')
+          BRANCH_NAME="db-update$now"
+          git branch "$BRANCH_NAME"
+          git checkout "$BRANCH_NAME"
+          git add -A
+          git commit -m "db: update manifests
```
ondrejbudai: What about putting `now` into the commit message? Is that a bad idea?

schutzbot/include_image_info.sh
```diff
@@ -0,0 +1,39 @@
...
+
+#remove ci-details-before-run
+mv ci-details-before-run /tmp
+rm -rf generated-image-infos
+
+# Amend the commit containing the db update
+git checkout $CI_COMMIT_BRANCH
+
+git config --local user.name "SchutzBot"
+git config --local user.email "schutzbot@redhat.com"
```
ondrejbudai: Same here.

schutzbot/include_image_info.sh
```diff
@@ -0,0 +1,39 @@
...
+
+git remote add origin2 https://schutzbot:"$SCHUTZBOT_GH_TOKEN"@github.com/schutzbot/manifest-db.git
```
ondrejbudai: `origin2` is not very descriptive, what about `upstream`, or `github`?

ondrejbudai (changes requested):
We are getting there, this is exciting! :heart_eyes:

README.md
```diff
@@ -0,0 +1,35 @@
+Manifest-db
+=======
+
+Stores all the  manifests and corresponding image-infos to test OSbuild.
+
+### Update workflow
+
+The update workflow starts with a github action that is manually triggered. This
+GH action is named `Update DB`. It will firstly clone the latests
```
ondrejbudai: `latests` -> `latest`

README.md
```diff
@@ -0,0 +1,35 @@
...
+`osbuild-composer` and copy all its manifests into the database. Secondly the
+action will create a commit and push this on the corresponding gitlab ci
+project
+(https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db).
+Thirdly, after the push is completed, the action triggers a special pipeline on
+the CI. This pipeline builds every manifests from the DB. Once building is done,
```
ondrejbudai: `manifests` -> `manifest`

README.md
```diff
@@ -0,0 +1,35 @@
...
+the pipeline starts the `schutzbot/include_image_info.sh` script. This script
+will download all the produced image-infos resulting from the builds and update
+the DB with them. Then it will amend the commit sent to the CI pipeline with
+these freshly builds image-info, push those commit to github on a branch and
```
ondrejbudai: `commit` -> `commits` (or maybe change everything to singular because there's only one commit? I'm actually not sure.)

ondrejbudai (changes requested):
Forgot the last commit, sorry!

tools/ci_import
lavocatt: Is it ignored ? The tool outputs a red cross if something went wrong including the error message.

tools/ci_import
lavocatt: Sometimes, for some jobs there's no artifacts to download, therefore we get a 404. The exception outputs some stuff anyway. At least in verbose mode. The tool should not crash here.

.github/workflows/auto_update.yml
lavocatt: So, the idea is to have the branch name to be every time different. So we don't run into the issue of recycling branches on gitlab and github. As for integrating it in the commit message, I think we can just rely on the commit date.

README.md
lavocatt: it should not be plural ?

tools/ci_import
ondrejbudai: But the pipeline is still marked as successful, isn't it?

lavocatt: Hmm right! Yes I should return 1 or something in that case.

README.md
lavocatt: yeah only one commit.

tools/ci_import
ondrejbudai: I don't think this is resolved.

tools/ci_import
```diff
@@ -0,0 +1,99 @@
...
+    jobs = pipeline.jobs.list(all=True)
+    for job in jobs:
+        if verbose:
+            print(f"Job {job.get_id()}:")
+        directory = path.join(tempfile.TemporaryDirectory().name, str(job.get_id()))
+        os.makedirs(directory, exist_ok=True)
+        filename = os.path.join(directory, "artifact.zip")
+        if verbose:
+            print(f"    downloading artifact.zip to {filename} for job {job.get_id()}")
+        try:
```
ondrejbudai: I think that this `try` block is too long. Can we put only the following line into it instead of the whole image updating logic that cannot throw `gitlab.exceptions.GitlabGetError`?
```python3
project.jobs.get(job.get_id()).artifacts(streamed=True, action=f.write)
```

ondrejbudai: An even better solution might be to put this block:

```python3
        filename = os.path.join(directory, "artifact.zip")
        if verbose:
            print(f"    downloading artifact.zip to {filename} for job {job.get_id()}")
        try:
            with open(filename, "wb") as f:
                project.jobs.get(job.get_id()).artifacts(streamed=True, action=f.write)
            z = zipfile.ZipFile(filename)
            z.extractall(directory)
            if verbose:
                print(f"    image information imported for {job.get_id()}")
            try:
                imis = os.listdir(os.path.join(directory, IMI_FOLDER))
            except FileNotFoundError:
                continue
```

into a separate function names something like `load_imis_from_gitlab_job`. It can return false if `artifact.zip` doesn't exist, or `IMI_FOLDER` inside the archive doesn't exist.

tools/ci_import
```diff
@@ -0,0 +1,99 @@
...
+def import_image_infos(pipeline, token, verbose=False):
+    success = True
+    gl = gitlab.Gitlab("https://gitlab.com", oauth_token=token)
+    project = gl.projects.get(MANIFEST_DB_PROJECT_ID)
+    pipeline = project.pipelines.get(pipeline)
+    jobs = pipeline.jobs.list(all=True)
+    for job in jobs:
+        if verbose:
+            print(f"Job {job.get_id()}:")
+        directory = path.join(tempfile.TemporaryDirectory().name, str(job.get_id()))
```
ondrejbudai: We usually try using `tempfile.TemporaryDirectory` as a context manager, so it gets cleaned up predictably. I think we can just use one tempdir for the whole `import_image_infos`.

See e.g. https://github.com/osbuild/osbuild-composer/blob/df22e8b6869b32018a10eb812bc37ff8878c3ab5/tools/deploy-qemu#L36 (there is also a ton of examples in `osbuild`)

tools/ci_import
```diff
@@ -0,0 +1,99 @@
...
+import tempfile
+from os import path
+
+import gitlab
+
+IMI_FOLDER = "generated-image-infos"
+MANIFEST_DB_PROJECT_ID = 36844106
+
+
+def import_image_infos(pipeline, token, verbose=False):
```
ondrejbudai: I would be in favor of splitting this method into smaller ones, it's on the verge of being hard to read. One suggestion that I made in a later comment is to factor out `load_imis_from_gitlab_job` method.

tools/ci_import
```diff
@@ -0,0 +1,99 @@
...
+                tc_data["image-info"] = imi_data
+                try:
+                    if verbose:
+                        print(f"        updating {test_case_path}:", end=" ")
+                    with open(test_case_path, "w", encoding="utf-8") as f:
+                        json.dump(tc_data, f, indent=2)
+                    if verbose:
+                        print("✅")
+                except Exception as e:  # pylint: disable=[broad-except]
+                    success = False
```
ondrejbudai: What about just not catching this exception at all? The script would then end with a non-zero exit code and print a stack trace which is helpful for debugging (otherwise, in a non-verbose mode, the script will just fail without any helpful error message).

tools/import-image-tests
ondrejbudai: This also appears unresolved.

schutzbot/include_image_info.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -euxo pipefail
+sudo dnf install gh -y
+sudo pip install python-gitlab
```
ondrejbudai: Have you tried `dnf install python3-gitlab`? Might be better to stick to OS dependencies for the sake of consistency.

schutzbot/include_image_info.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Login to GH
+echo "${SCHUTZBOT_GH_TOKEN}" > /tmp/secret.txt
```
ondrejbudai: What about simplifying this with a pipe?

```
echo "${SCHUTZBOT_GH_TOKEN}" | gh auth login --with-token
```

schutzbot/include_image_info.sh
```diff
@@ -0,0 +1,39 @@
...
+gh auth login --with-token < /tmp/secret.txt
+
+# import the image info from the current build
+./tools/ci_import --pipeline-id "$CI_PIPELINE_ID" --token  ${GITLAB_TOKEN} --verbose
```
ondrejbudai: Double-space before `${GITLAB_TOKEN}`, I would also quote it just to be extra-safe. (shellcheck would also be nice in this repository)

schutzbot/include_image_info.sh
```diff
@@ -0,0 +1,39 @@
...
+
+#remove ci-details-before-run
+mv ci-details-before-run /tmp
+rm -rf generated-image-infos
+
+# Amend the commit containing the db update
+git checkout $CI_COMMIT_BRANCH
```
ondrejbudai: quoting once again

.github/workflows/auto_update.yml
```diff
@@ -0,0 +1,49 @@
...
+          git add -A
+          git commit -m "db: automatic update of manifests."
+
+          mkdir -p ~/.ssh
+          echo "${IMAGEBUILDER_BOT_GITLAB_SSH_KEY}" > ~/.ssh/id_rsa
+          chmod 400 ~/.ssh/id_rsa
+          touch ~/.ssh/known_hosts
+          ssh-keyscan -t rsa gitlab.com >> ~/.ssh/known_hosts
+          git remote add ci git@gitlab.com:redhat/services/products/image-builder/ci/manifest-db.git
+          git push -f ci
```
ondrejbudai: Doesn't this create a "regression" gitlab pipeline? Maybe a `ci.skip` would be nice here?

https://docs.gitlab.com/ee/user/project/push_options.html#push-options-for-gitlab-cicd

README.md
```diff
@@ -0,0 +1,29 @@
+Manifest-db
+=======
+
+Stores all the  manifests and corresponding image-infos to test OSbuild.
```
ondrejbudai: - Double white-space before `manifests`.
- OSbuild => OSBuild, see https://www.osbuild.org/

README.md
ondrejbudai: Nah, `every` is used before a singular noun, see https://dictionary.cambridge.org/dictionary/english/every

README.md
```diff
@@ -0,0 +1,29 @@
...
+`osbuild-composer` and copy all its manifests into the database. Secondly the
+action will create a commit and push this on the corresponding gitlab ci
+project
+(https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db).
+Thirdly, after the push is completed, the action triggers a special pipeline on
+the CI. This pipeline builds every manifests from the DB. Once building is done,
+the pipeline starts the `schutzbot/include_image_info.sh` script. This script
+will download all the produced image-infos resulting from the builds and update
+the DB with them. Then it will amend the commit sent to the CI pipeline with
+these freshly builds image-info, push this commit to github on a branch and
```
ondrejbudai: `these freshly builds image-info` => `these freshly built image-infos`

ondrejbudai (changes requested):
:heart:

I think that `ci_import` would benefit from a clean-up and I found some whitespace issues and typos, otherwise this is very nice.

tools/ci_import
lavocatt: I've done an other kind of refactor, but keeping the idea of simplifying this part.

tools/ci_import
```diff
@@ -0,0 +1,122 @@
...
+        print(f"    no image information for {job.get_id()}")
+        return []
+
+
+def update_db(image_info_path):
+    """
+    Take a downloaded image info and write its content into the corresponding
+    test case in the database.
+    """
+    image_info = image_info_path.split("/")[-1]
```
ondrejbudai: https://docs.python.org/3/library/os.path.html#os.path.basename would be slightly better.

ondrejbudai (approved):
Love the refactor, I think this is ready to go.

lavocatt:
I'm waiting for the new test to finish before merging. If everything is fine, this pipeline https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/610351365 will open a PR here. If it does, we can merge.

db-ignore
```diff
@@ -0,0 +1,10 @@
+centos_8-x86_64-image_installer-boot
+centos_8-x86_64-image_installer_with_users-boot
+centos_9-x86_64-image_installer-boot
+centos_9-x86_64-image_installer_with_users-boot
+rhel_85-x86_64-image_installer-boot
+rhel_85-x86_64-image_installer_with_users-boot
+rhel_86-x86_64-image_installer-boot
+rhel_86-x86_64-image_installer_with_users-boot
+rhel_90-x86_64-image_installer-boot
+rhel_90-x86_64-image_installer_with_users-boot
```
gicmo, line 10: Why are we ignoring those? Can we add a comment in the file about it? Can do it in follow up.

tools/import-image-tests
```diff
@@ -63,9 +63,16 @@ def load_test(name, fullpath) -> dict:
...
-def export_test(test, target):
+def export_test(test, target, manifest_only):
     name = test["id"]
     fullpath = os.path.join(target, name + ".json")
+    if manifest_only and os.path.exists(fullpath):
+        # load existing
+        with open(fullpath, "r", encoding="utf-8") as f:
+            base = json.load(f)
+        # only replace the manifest in the base
+        base["manifest"] = test["manifest"]
```
gicmo, line 74: What else is could be in the file? The image info? Wouldn't the manifest and image-info be out of sync then?

lavocatt: So in each database file you have:
* the manifest
* the image-info

The update process requires to first update the manifest, then to send that to CI, at the end of the CI job, the commit is amended to integrate the image-info part of it. So it's "out of sync" only on CI, and when it lands as a PR they must be synced. And I think that's part of the review process to ensure that a change in a manifest should result in a change in the image-info side of the DB.

db-ignore
lavocatt: The resulting image-info of theses manifests aren't explorable by the image-info tool. That's why they are ignored. The commit bringing these explains this reason but yes, having some comments somewhere, either in the README or in the file directly might be nice. I'll create a ticket to handle it in a follow up.

tools/import-image-tests
lavocatt: though we could imagine a process in a follow up that tries to ensure that every changed manifest did result in a changed image-info. But it'll need some thinking.

gicmo: So `image-info` on an install fails? or is just not very useable? I should know, but I forgot ;)

db-ignore
gicmo: So image-info on an install fails? or is just not very useable? I should know, but I forgot ;)

lavocatt: Some of them are just failing. Like they make the image-info tool crash.

lavocatt: I think that would be nice to solve that, but that was out of the scope of this contribution I think :)

ondrejbudai:
The last commit can be squashed into the first one.

PR #30, opened 2022-08-17
Update the CI suite to the latest distros

lavocatt:
This PR updates the CI suite to the latest distros. By doing so it also tackle the issue of EOL distributions. As we are dropping some, we are therefore not compiling anymore their image-info during the automated process. Leaving an out of sync DB where the input might manifest have changed but not its resulting image-info. Now the import procedure only imports the manifests from the distributions declared in the gitlab yaml file for CI. The rest of them will be left not updated.

The .gitlab-ci.yml file is parametrized to follow the test on osbuild.
A test automated update is currently running: https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/616874644

The resulting PR of this pipeline should contain only image-info updates for the distros found in the gitlab-ci.yml file. If this criteria is not met the PR can't be merged:

- [x] #32 contains modifications only to the manifest-db entries related to archs and distros handled by CI (only consider last commit)


lavocatt:
It well created the PR see https://github.com/osbuild/manifest-db/pull/31 but, I'm not happy with the result.
Cleaning the image-info also alters ppce archs, and I don't want to mess with them.
I think a better strategy is to only import from composer the manifests that will be recomputed during the CI. The tool could easily know that by accessing the gitlab ci yml file.

lavocatt:
So I figured out the best way to treat the inconsistency issue:

We only import from composer the manifests that can be built by CI with the automated update.
This way, we don't mess with what we can't build. If it's gets outdated, well, that shouldn't be that bad either because, we will keep the list of distros on par with what osbuild tests. And it lets us the ability to update parts of the DB outside the automatic process without the fear of these updates being erased by the bot.

lavocatt:
On the pull request opened by the bot #32, the only items changed in the DB are concerning archs and distros handled by the ci. So that's a success.

tools/import-image-tests
```diff
@@ -147,15 +157,43 @@ def main():
...
+                    flavor = machine_id.split("/")[-1]
+                    arch = flavor.split("-")[-1]
+                    distro = ""
+                    if "fedora" in flavor:
+                        distro = f"fedora-{flavor.split('-')[1]}"
+                    elif "centos" in flavor:
+                        distro = f"centos-{flavor.split('-')[2]}"
+                    elif "rhel" in flavor:
+                        distro = f"rhel-{flavor.split('-')[1].replace('.', '')}"
+                    print(f"    - {distro} {arch}")
```
supakeen: Does it make sense to have `distro` be allowed to be empty?

tools/import-image-tests
```diff
@@ -49,6 +50,10 @@ def load_test(name, fullpath) -> dict:
     if not req:
         raise ValueError("no compose-request")
 
+    distros, archs = list(map(list, zip(*filters)))
```
supakeen: I believe just `distros, archs = zip(*filters))` is enough, no need to turn things into intermediate lists.

tools/import-image-tests
```diff
@@ -37,7 +38,7 @@ def list_tests(source) -> Iterator[Tuple[str, str]]:
         yield name, fullpath
 
 
-def load_test(name, fullpath) -> dict:
+def load_test(name, fullpath, filters) -> Optional[dict]:
```
supakeen: If you do the typing note that we need to support 3.6 so we can't use default type subscription (that was a 3.8 (?) feature). Use `Optional[Dict]` and *preferably* also type the `Dict` generic as mypy will error on it in strict mode: `Optional[Dict[str, Any]]`. And honestly, I'd just make it `Dict[str, Any]` and return an empty dict instead of None :)

tools/import-image-tests
```diff
@@ -147,15 +157,43 @@ def main():
     if args.db_ignore:
         with open(args.db_ignore, "r", encoding="utf-8") as f:
             db_ignore = f.read().splitlines()
+        print("Filtering manifests, ignoring:")
```
supakeen: I'd add the print to where the actual ignore takes place (line 190).

tools/import-image-tests
```diff
@@ -147,15 +157,43 @@ def main():
...
+                    arch = flavor.split("-")[-1]
+                    distro = ""
+                    if "fedora" in flavor:
+                        distro = f"fedora-{flavor.split('-')[1]}"
+                    elif "centos" in flavor:
+                        distro = f"centos-{flavor.split('-')[2]}"
+                    elif "rhel" in flavor:
+                        distro = f"rhel-{flavor.split('-')[1].replace('.', '')}"
+                    print(f"    - {distro} {arch}")
+                    filters.append((distro,arch))
```
supakeen: I'd probably do this as a `dict[str, list[str]]` but that's just me running out of things to say ;)

tools/import-image-tests
lavocatt: Not really, I forgot to crash on the else case! thanks

tools/import-image-tests
lavocatt: That might be better indeed.

tools/import-image-tests
lavocatt: Not at all :D

thozza:
I wanted to ask about the removed manifests which can't be inspected by `image-info`. The previous image tests were at least building such manifests and then not comparing image-info report if there is nothing to compare. I would argue that it is still beneficial to test that osbuild can build such manifests. However I acknowledge that this does not cover the main goal of testing that osbuild produces the same image for the same manifest.

lavocatt:
> I wanted to ask about the removed manifests which can't be inspected by `image-info`. The previous image tests were at least building such manifests and then not comparing image-info report if there is nothing to compare. I would argue that it is still beneficial to test that osbuild can build such manifests. However I acknowledge that this does not cover the main goal of testing that osbuild produces the same image for the same manifest.

That's quite an interesting point. Maybe we can come up with something, maybe on osbuild's side or here in a special kind of CI workflow where we run every builds just to see if it makes it without comparing ? @gicmo what do you tihnk about this ?

thozza (approved):
👍

tools/import-image-tests
```diff
@@ -63,40 +60,78 @@ def load_test(name, fullpath) -> dict:
...
+    if verbosity < 1:
+        return
+
+    if verbosity == 1:
+        print(f"{RED}F{RESET}", end="", flush=True)
+    else:
+        print(f"{RED}FAIL {RESET} {name}: {reason}")
+
+
+def parse_distro(ci_file: Union[str, pathlib.Path]) -> Dict[str, List[str]]:
```
gicmo, line 106: Nit: not really a `distro` what you are parsing here (nor is `distro` the result).

tools/import-image-tests
```diff
@@ -63,40 +60,78 @@ def load_test(name, fullpath) -> dict:
...
+    filters: Dict[str, List[str]] = {}
+    if ci_file:
```
gicmo, line 108: Just a side node, I think an early return makes the code more readable because you can get rid of a level of indention.

tools/import-image-tests
```diff
@@ -63,40 +60,78 @@ def load_test(name, fullpath) -> dict:
...
+                    elif "centos" in flavor:
+                        distro = f"centos-{flavor.split('-')[2]}"
+                    elif "rhel" in flavor:
+                        distro = f"rhel-{flavor.split('-')[1].replace('.', '')}"
+                    else:
+                        raise ValueError(f"unknown distribution {flavor}")
+                    if distro in filters:
+                        filters[distro].append(arch)
+                    else:
+                        filters[distro] = [arch]
```
gicmo, line 128: Could use `collections.defaultdict(list)` to remove the `if`.

tools/import-image-tests
```diff
@@ -63,40 +60,78 @@ def load_test(name, fullpath) -> dict:
...
+    return filters
+
+def parse_db_ignore(db_ignore_file: Union[str, pathlib.Path]) -> List[str]:
+    with open(db_ignore_file, "r", encoding="utf-8") as f:
+        return f.read().splitlines()
```
gicmo, line 133: The `default` for `"--filter-with-ci-distros"` is `None` but this function does not handle that.

tools/import-image-tests
```diff
@@ -6,26 +6,20 @@ Import Composer image tests
...
-    comps = name.split("-")
-    distro = comps[0]
-    arch = comps[1]
-    name = comps[2]
-    extra = comps[3] if len(comps) > 3 else None
-    return distro, arch, name, extra
-
-
-def list_tests(source) -> Iterator[Tuple[str, str]]:
+def list_tests(source: Union[str, pathlib.Path]) -> Iterator[Tuple[str, str]]:
```
gicmo, line 22: Heads up, using `Iterator` is `Deprecated since version 3.9`. It is a bit annoying. I think `typing.Generator` is the safest replacement.

tools/import-image-tests
```diff
@@ -37,7 +31,7 @@ def list_tests(source) -> Iterator[Tuple[str, str]]:
         yield name, fullpath
 
 
-def load_test(name, fullpath) -> dict:
+def load_test(name: str, fullpath: Union[str, pathlib.Path], filters: Dict[str, List[str]]) -> Dict[str, Any]:
```
gicmo, line 34: Maybe we need to define some aliases for some of the types (like we do in osbuild for the Path related things), because this is now really hard to read with all the type parameters.

.github/workflows/auto_update.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Update DB
         run: |
-          ./manifest-db/tools/import-image-tests composer/test/data/manifests/ manifest-db/manifest-db --manifest-only --db-ignore=manifest-db/db-ignore --verbose
+          ./manifest-db/tools/import-image-tests composer/test/data/manifests/ manifest-db/manifest-db --manifest-only --db-ignore=manifest-db/db-ignore --filter-with-ci-distros=manifest-db/.gitlab-ci.yml --verbose
```
gicmo, line 27: Random observation: `--verbose`, i.e. verbosity 1, is the default anyway.

tools/import-image-tests
```diff
@@ -63,40 +60,78 @@ def load_test(name, fullpath) -> dict:
...
+                for machine_id in runner["RUNNER"]:
+                    flavor = machine_id.split("/")[-1]
+                    arch = flavor.split("-")[-1]
+                    distro = ""
+                    if "fedora" in flavor:
+                        distro = f"fedora-{flavor.split('-')[1]}"
+                    elif "centos" in flavor:
+                        distro = f"centos-{flavor.split('-')[2]}"
+                    elif "rhel" in flavor:
+                        distro = f"rhel-{flavor.split('-')[1].replace('.', '')}"
```
gicmo, line 122: Small side note: this might be a bit hard to maintain without any documentation or examples.

gicmo (commented):
Sorry, this is a bit late, but I promised to have a look so I did.

PR #38, opened 2022-08-24
Fix the SELinux mismatch

lavocatt:
The issues are: 
* I have missing packages in my setup and that leads to SELinux mismatch between what I'm computing on this CI and OSBuild's CI.
* The `which` command used alongside `matchpathcon` does not give the same result when its called from root or user context. And therefore we don't have access to the proper label for the `image-info` command.

The first issue is fixed by by copying the `deploy.sh` setup procedure from `OSBuild`.
The second issue is fixed by specifying the path to access the `osbuild` executable when calling `matchpathcon`. 

Now `deploy.sh` uses the composer-tests package to have the base configuration even though we are not using the tests coming from this package. A probable next step would be to create it's own package definition for manifest-db and to install it the way we are installing composer-tests package.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,7 @@ Regression:
   extends: .terraform
   script:
     - schutzbot/deploy.sh
-    - sudo schutzbot/selinux-context.sh
+    - schutzbot/selinux-context.sh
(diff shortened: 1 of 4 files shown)


achilleas-k (approved):
LGTM

PR #42, opened 2022-08-30
tool/update_tool: new tool to explore db updates

lavocatt:
Db updates are tricky to review. This tools brings some utility
functions to try to simplify the process. It gives you access to 3
commands: Ls, Diff and Report

List the changed DB entry and know if the manifest or the image info was
updated easily. The tool also ignores by default the volatiles UUID so
you are not polluted by that during the review process.

Diff the changed files, either all of them one after the other or a
specific set given by the user. It will open a difftool (vimdiff by
default) to diff either the manifest or the image-info (or both if they
both changed).

Report the change as an ascii table with color codes (by default) or as
a github TODO list to be integrated within the github PR.

### Terminal report

Peek updatetool_report

### Github report

Peek updatetool_report_gh

## Listing changed DB entries

Peek updatetool_ls

## Diff all or a list of files

Peek updatetool_diff

### DIFF
--- a/tools/update_tool
+++ b/tools/update_tool
@@ -1,0 +1,424 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import os
+import sys
+import tempfile
+import subprocess
+from abc import ABC, abstractmethod
+import columnify
+
+RESET = "\033[0m"
+GREEN = "\033[32m"
+BOLD = "\033[1m"
+RED = "\033[31m"
+YELLOW = "\033[33m"
+STRIKE = "\033[9m"
+
+
+class Report(ABC):
+
+    def __init__(self):
+        self.header()
+
+    @abstractmethod
+    def header(self):
+        pass
+
+    @abstractmethod
+    def row(self, name, mstatus, istatus):
+        pass
+
+
+class CliReport(Report):
+
+    status_map = {"added": f"{BOLD}{GREEN}",
+                  "updated": f"{GREEN}",
+                  "removed": f"{STRIKE}{RED}",
+                  "unchanged": f"{YELLOW}"}
+
+    def __init__(self):
+        self.terminal_width = os.get_terminal_size().columns
+        section1 = 0.7
+        section2 = 0.15
+        section3 = 0.15
+        self.nwidth = int(self.terminal_width*section1)-5
+        self.mwidth = int(self.terminal_width*section2)-2
+        self.iwidth = int(self.terminal_width*section3)-1
+        super().__init__()
+
+    @classmethod
+    def truncate(cls, s, l):
+        if len(s) > l:
+            s = f"{s[0:l-3]}..."
+        return s
+
+    @classmethod
+    def adjust(cls, s, l):
+        return cls.truncate(s, l).ljust(l)
+
+    def header(self):
+        print("| ", "DB entry".ljust(self.nwidth), end="| ")
+        print(self.adjust("manifest", self.mwidth), end="| ")
+        print(self.adjust("image-info", self.iwidth), end="|\n")
+
+        print(f"|-{'-'*(self.nwidth+1)}", end="|-")
+        print(f"{'-'*self.mwidth}", end="|-")
+        print(f"{'-'*self.iwidth}", end="|\n")
+
+    def row(self, name, mstatus, istatus):
+        name = self.truncate(name, self.nwidth)
+        ms = CliReport.status_map[mstatus]
+        ims = CliReport.status_map[istatus]
+        print("| ", f"{name}".ljust(self.nwidth), end="| ")
+        print(f"{ms}{self.adjust(mstatus,self.mwidth)}{RESET}", end="| ")
+        print(f"{ims}{self.adjust(istatus,self.iwidth)}{RESET}", end="|\n")
+
+        print(f"|-{'-'*(self.nwidth+1)}", end="|-")
+        print(f"{'-'*self.mwidth}", end="|-")
+        print(f"{'-'*self.iwidth}", end="|\n")
+
+
+class GHReport(Report):
+
+    status_map = {"added": "🆕",
+                  "updated": "🔁",
+                  "removed": "🗑️",
+                  "unchanged": "💾"}
+
+    def header(self):
+        print("Please review each DB entry")
+
+    def row(self, name, mstatus, istatus):
+        print(name, end=": ")
+        print(f"manifest: {GHReport.status_map[mstatus]}{mstatus}", end=", ")
+        print(f"image-info:{GHReport.status_map[istatus]}{istatus}.")
+
+
+def compare(obj1, obj2):
+    """
+    Compare two json objects and return as a string the conclusion of the diff.
+    Obj1 is the version of Obj2 back in time. So when obj2 is not and obj1 is,
+    we can say that something was removed and vice versa.
+    """
+    if obj1 and not obj2:
+        return "removed"
+    if not obj1 and obj2:
+        return "added"
+    if obj1 != obj2:
+        return "updated"
+    return "unchanged"
+
+
+def list_changed_files():
+    """
+    Go through the diff between this commit and the last one and list all the
+    files in there. An update is only valid if it contains only the DB part that
+    is updates anyway.
+    """
+    try:
+        command = subprocess.run(
+            ["git", "diff-tree",  "--no-commit-id", "--name-only", "HEAD", "-r"],
+            capture_output=True,
+            check=True)
+        return command.stdout.decode("utf-8").split("\n")
+    except subprocess.CalledProcessError:
+        return None
+
+
+def get_db_entry_at_previous_version(file):
+    """
+    Retrieves a version of a DB file from the previous commit state.
+    """
+    if not file:
+        return None
+    try:
+        command = subprocess.run(
+            ["git", "show", f"HEAD^:{file}"],
+            capture_output=True,
+            check=True)
+        return json.loads(command.stdout)
+    except subprocess.CalledProcessError:
+        return None
+    return None
+
+
+def get_db_entry_at_current_version(file):
+    """
+    Retrieves a version of a DB file at the current state of the DB
+    """
+    if not file:
+        return None
+    with open(file, "r", encoding="utf-8") as f:
+        return json.load(f)
+
+
+def filter_fn(imi):
+    """
+    Filter specific fields in the image-info that can't be compared
+    together.
+    """
+    def lvm2(imi):
+        """
+        LVM2 partitions have a UUID that is not fixed. Replace the value
+        upon comparison time
+        """
+        partitions = imi.get("partitions")
+        if partitions:
+            for partition in partitions:
+                if partition.get("fstype") == "LVM2_member":
+                    partition["uuid"] = "2022-07-01-fixed-uuid"
+        return imi
+
+    def iso(imi):
+        """
+        For isos, the partition UUID is the date of the build. Replace
+        that with a fixed one for the comparison.
+        """
+        if "image-format" in imi and "type" in imi["image-format"] and imi["image-format"]["type"] == "raw":
+            if "partitions" in imi:
+                for partition in imi["partitions"]:
+                    if "fstype" in partition:
+                        if partition["fstype"] == "iso9660":
+                            partition["uuid"] = "2022-07-01-fixed-uuid"
+        return imi
+    return iso(lvm2(imi))
+
+
+def diff(what, json1, json2, tool="vimdiff"):
+    """
+    Starts a diff between two json objects with a specific tool to do so. Writes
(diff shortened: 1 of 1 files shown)


ondrejbudai:
May I ask you for a quick example of the output? That would be very valuable to everyone having a look at this PR. :)

lavocatt:
> May I ask you for a quick example of the output? That would be very valuable to everyone having a look at this PR. :)

Good idea! I'll record my screen using it with peek and include the gifs inside the PR description and in the README also :)

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+
+def ls_command(args):
+    """
+    Lists all the changed files between this and the previous commit and puts
+    some icons for the user to know what changed and what not
+    """
+    files = []
+    lst = list_changed_files()
+    for i, file in enumerate(lst):
+        filename = file.split("/")[-1].split(".")[0]
```
achilleas-k: `os.path` has functions for doing these specific to paths.
see `os.path.basename()` and `os.path.splitext()`.

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+        prev = get_db_entry_at_previous_version(file)
+        curr = get_db_entry_at_current_version(file)
+
+        if not (prev or curr):
+            continue
```
achilleas-k, line 262: If one of the two doesn't exist, couldn't we just shortcut the function to say something was introduced (if prev is None) or something was removed (if curr is None)?

achilleas-k (commented):
Added some comments in the `diff` function that apply to all places where the same code and logic is used.

tools/update_tool
lavocatt: Thanks for review! Yes that's a good idea. I think the logic is already kind of handled, because when you are in an added or removed state you'll get the info. Maybe you'd rather want the code to avoid opening you a diff in these two cases ?
At the beginning I wanted to only open the diff tool on the updated state.

We could instead of having:

```
        if manifest_status != "unchanged":
            diff("manifest", prev_manifest, curr_manifest, tool=args.manifest_diff_tool)

        if image_info_status != "unchanged":
            diff("image-info", prev_imi, curr_imi, tool=args.image_info_diff_tool)
```
            
have:
      
```
        if manifest_status == "updated":
            diff("manifest", prev_manifest, curr_manifest, tool=args.manifest_diff_tool)

        if image_info_status == "updated":
            diff("image-info", prev_imi, curr_imi, tool=args.image_info_diff_tool)
``` 
 
 What do you think ?

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+        action="store_true",
+        default=False,
+        help="Disable UUID masking"
+    )
+    subparsers = parser.add_subparsers(
+        title="command",
+        required=True,
+        dest='command',
+        help='Command to execute')
+    parser_report = subparsers.add_parser('report')
```
ondrejbudai: I think this line should be after `# Report command`

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+    parser_diff = subparsers.add_parser('diff')
+    parser_diff.add_argument(
+        "--manifest-diff-tool",
+        default="vimdiff",
+        help="tool to diff the manifests"
+    )
+    parser_diff.add_argument(
+        "--image-info-diff-tool",
+        default="vimdiff",
+        help="tool to diff the manifests"
```
ondrejbudai: Shouldn't this `help` be `tool to diff the image infos`?

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+    # Report command
+    parser_report.add_argument(
+        "--github-markdown",
+        action="store_true",
+        default=False,
+        help="Create a TODO list for github to ask a reviewer to check everything"
+    )
+
+    # Diff command
+    parser_diff = subparsers.add_parser('diff')
```
ondrejbudai: I think we should have `description=` for all subparsers so it's easier to find out what the subcommands do,

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+    l = 77
+    if gh:
+        l = 0
+        end = ", "
+        prefix = "- [ ] "
+
+    print(f"{prefix}{text}".ljust(l), end=end)
+
+
+def green_adjusted(text, bold=False, strike=False, l=18, gh=False, last=False):
```
ondrejbudai: `green_adjusted` is basically the same thing as `red_adjusted`. Btw, have you considered using something like `jinja` to do the formatting? I think a declarative template might be simpler that doing the formatting imperatively,

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+            continue
+
+        prev_manifest, prev_imi = extract(prev, do_filter=not args.no_filter)
+        curr_manifest, curr_imi = extract(curr, do_filter=not args.no_filter)
+
+        manifest_status = compare(prev_manifest, curr_manifest)
+        image_info_status = compare(prev_imi, curr_imi)
+
+        if manifest_status != "unchanged" or image_info_status != "unchanged":
+            ms = get_symbol(compare(prev_manifest, curr_manifest))
```
ondrejbudai:
```suggestion
            ms = get_symbol(manifest_status)
```

Same for the following line.

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+    else:
+        print(columnify.columnify(
+            items=files,
+            line_width=os.get_terminal_size().columns))
+
+
+def diff_command(args):
+    """
+    Diffs either some files or all the files with the difftool provided by the
+    user. Default tool is vimdiff
```
ondrejbudai: A small nitpick: `diff_command` doesn't know anything about the default diff tool, that's set higher.

tools/update_tool
```diff
@@ -0,0 +1,466 @@
+#!/bin/env python
+
+import argparse
+import json
+import os
+import sys
+import tempfile
+import shlex
```
ondrejbudai: I suggest removing `shlex`, both its uses can be replaced by a simple array.

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+        end = "|"
+        if bold:
+            bo = BOLD
+        st = ""
+        if strike:
+            st = STRIKE
+        print(f"{RED}{bo}{st}{text.ljust(l)}{RESET}", end=end)
+
+
+def print_header(gh):
```
ondrejbudai: I wonder if we could change this method to the form of:

```python3
def print_header(gh):
    if gh:
         # github-specific formatting
    else:

         # terminal-specific formatting
```

ondrejbudai (changes requested):
Thanks a lot for the demos, they look really sleek. Some suggestions inline to make the code nicer.

@supakeen May I ask for your opinion on using some kind of templating? I'm still not sure if it improved readability, but I'm just not a fan of formatting text imperatively.

tools/update_tool
lavocatt: I'm really glad you catch my inattention errors :smile:

tools/update_tool
lavocatt: Yes, that chunk of code is not my favourite here. You can see that I went back and forth many times and that are leftovers from the first ideas. That needs a cleanup.

supakeen:
@ondrejbudai I'll take a look later this morning, thanks.

tools/update_tool
```diff
@@ -0,0 +1,466 @@
+#!/bin/env python
```
supakeen: Please point to `#!/usr/bin/env python3`, the `python3` bit being the important one here, I believe `/usr/bin/env` is also the POSIX standard location.

```suggestion
#!/usr/bin/env python3
```

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+        print("-"*116)
+
+
+def compare(obj1, obj2):
+    """
+    Compare two json objects and return as a string the conclusion of the diff.
+    Obj1 is the version of Obj2 back in time. So when obj2 is not and obj1 is,
+    we can say that something was removed and vice versa.
+    """
+    if obj1 and not obj2:
```
supakeen, line 105: Bit of a nitpick, but is an empty `obj1` (say: `{}`) the same as `obj1` being `None`? If it isn't it needs an identity check for the latter in these conditionals.

tools/update_tool
supakeen: I am irked by the formatting code as well, I wouldn't suggest going down the actual templating engine path but I *would* suggest moving all the text formatting functions into a class, having the class encapsulate the `is_github` boolean.

Most of these formatting functions can also be 'cleaned up' but that's just an opinion; e.g. I'd write them more in this idiom:

```
def start_row(text, is_github):
    min_width = 77

    end = ", " if is_github else "|"
    pre = "- [ ] " if is_github else ""

    print(f"{pre}{text:<{min_width}}{end}", end="")
```

You can then move shared bits into the format class (e.g; `end = "|"`) is repeated a few times, so are others.

tools/update_tool
```diff
@@ -0,0 +1,466 @@
...
+        if do_filter:
+            imi = filter_fn(db_entry["image-info"])
+        else:
+            imi = db_entry["image-info"]
+    if db_entry["manifest"]:
+        manifest = db_entry["manifest"]
+    return manifest, imi
+
+
+def get_symbol(label):
```
supakeen: Minor: this function could be a dictionary.

tools/update_tool
lavocatt: That's a good one! Thanks for this. I'd say, they should be considered different, something changed in the DB, so, we should highlight that.

tools/update_tool
lavocatt: I kinda like having this, it's handy to just copy paste the command and split it without doing it manually, but if you have a strong opinion against shlex, I can remove it.

supakeen: The problem would be a mismatch between how `shlex` understands the command and how the developer understands the command. Writing them as list literals would stop me from having to think about 'wait how does `shlex` split things again?'.

tools/update_tool
```diff
@@ -0,0 +1,439 @@
...
+    def header(self):
+        print("| ", "DB entry".ljust(self.nwidth), end="| ")
+        print(self.adjust("manifest", self.mwidth), end="| ")
+        print(self.adjust("image-info", self.iwidth), end="|\n")
+
+        print(f"|-{'-'*(self.nwidth+1)}", end="|-")
+        print(f"{'-'*self.mwidth}", end="|-")
+        print(f"{'-'*self.iwidth}", end="|\n")
+
+    status_map = {"added": f"{BOLD}{GREEN}",
```
supakeen: Very minor but I like having my class attributes near the top of the class (listed first).

supakeen (commented):
I like the way the formatting is structured now much more.

tools/update_tool
```diff
@@ -0,0 +1,439 @@
...
+        self.iwidth = int(self.terminal_width*section3)-1
+        super().__init__()
+
+    @classmethod
+    def truncate(cls, s, l):
+        if len(s) > l:
+            s = f"{s[0:l-3]}..."
+        return s
+
+    def adjust(self, s, l):
```
ondrejbudai: This can be `@classmethod` as well, right?

tools/update_tool
ondrejbudai: This is still unresolved.

ondrejbudai (commented):
Looks much better, thanks! There are still some unresolved comments but they should be easy to fix. :)

tools/update_tool
lavocatt: Sorry, I forgot out of inattention some of the comments :/ Taking care of this now.

tools/update_tool
lavocatt: I see what you mean. Alright, you're both against, I'm following the majority, even though I dislike splitting my commands manually. I'll make myself a vim function for that :smile:

tools/update_tool
lavocatt: @achilleas-k, I'll mark this as resolved, but feel free to reopen it.

tools/update_tool
```diff
@@ -0,0 +1,439 @@
...
+        if do_filter:
+            imi = filter_fn(db_entry["image-info"])
+        else:
+            imi = db_entry["image-info"]
+    if db_entry["manifest"]:
+        manifest = db_entry["manifest"]
+    return manifest, imi
+
+
+def get_symbol(label):
```
ondrejbudai: Oh, I didn't see this before. I think this is unused now, right? Can you remove it?

ondrejbudai (changes requested):
One forgotten function, otherwise this is fine. :)

tools/update_tool
lavocatt: Thanks for catching that!

ondrejbudai (approved):

PR #53, opened 2022-09-26
Image info refactor

lavocatt:
This PR brings a complete refactor of the `image-info` tool.

**Goals:**
- Easily extensible architecture
- Code base reusable to create a `difftool` for image-info

**Tasks**
- [x] refactor
- [x] unit tests to validate no regression in the produced JSON while loading an existing image-info JSON
- [x] regression tests to validate no regression in the produced JSON while analyzing builds
- [ ] document

**The new architecture**:

Instead of having everything a single script, I aimed for a python package definition with core functionalities and extensible ones.
The package is divided like this:

- `main_cli`: entry points
- `core`: central elements to the target framework
- `report`: definition of the report elements used to build reports. That's where target specific elements are defined.
- `common`: definition of report elements that are common to every target.
- `utils`: utilities functions

`core/target.py ` is the main framework class. The idea behind this class is that, when you have something to analyze with \image-info` target will discover which kind of target this something is about, load the correct subclass and return it to you. You can either ask Target to load from a JSON file or from a standard OSBuild result. The workflow is a bit different for the two usecases.

When you load a Target from a JSON file, you endup having a correct target for the loading JSON, in the sense that it will be either an ImageTarget, an Ostree target or a DirTarget. But you won't be able to invoke `inspect()` on it since, well, there no target to inspect.

When you load a Target from an OSBuild result, then you endup with one you should invoke `inspect()` on, if you want a report ofc.

Then the following steps will occur:
- Depending of the subclass some preliminary work will be executed:
    - for a DirTarget, its nothing
    - for an OstreeTarget, it's adding a couple of specific ReportElements associated with Ostree
    - for an ImageTarget, it's finding the partitions, mounting them and adding a few ReportElement specific to an ImageTarget
    - for a CompressedTarget, it's unziping + One of the above
    - for a TarballTarget, it's unziping + an ImageTarget
- Then the Commons are added.

**The commons**
Every class inheriting from `Common` and being defined in a `$file.py` under the `common` folder is guaranteed to be dynamically loaded. When it's time for the commons to be added to a report, the Target simply loop to all the discovered ones and present them a `device` or a `json` to analyze.

This ticks the goal "`image-info` easy to extend". If you want to add a new subsection applicable to every kind of Target we support, you just have to add a new class inheriting from Common under the common directory.

**The partition mounting of an ImageTarget**

This bit is the one that received the most modifications from the legacy `image-info`. Bits of the mounting are scattered between the `core/filesystem.py`, `report/partition.py` and `report/image.py` files.

To help see more clearly the interactions with all the classes, here a class diagram:

```mermaid
classDiagram
    GenericPartition~ReportElement~ <|-- ImagePartition
    FileSystemFactory <|-- ImagePartition
    GenericPartition <|-- LvmVolume
    ImagePartition <|-- LvmPartition
    LvmPartition "1" *-- "1..*" LvmVolume: discover_lvm()
    FileSystemFactory "1" *-- "1..*" FileSystem
    ImagePartition "1" *-- "1" LvmPartition: to_lvm_partition()
    FileSystem "1" *-- "0..*" FstabEntry
    PartitionTable~ReportElement~ "1" *-- "1..*" GenericPartition
    PartitionTable -- FileSystemMounter
    FileSystemMounter "1" o-- "1..*" FileSystem

class FileSystem{
    List~FstabEntry~ fstab()
    mount_root()
    mount_at()
}

class FileSystemFactory{
    List~FileSystem~ fsystems()
}

class PartitionTable~ReportElement~{
    from_device()
    bool has_partition_table()
    mount()
}

class FileSystemMounter{
    __init__(List~FileSystem~)
    mount_all()
}
```


lavocatt:
Currently testing the update with this image-info version https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/662654995

thozza:
Just a few preliminary observations from my side:

- I would welcome an executable script allowing me to execute `image-info` from a git repo checkout, without having to install it. There currently does not seem to be any easy and obvious way to do it without creating it myself.
- I would propose to keep a similar structure as used by `osbuild` stages for files in `image_info/common` and split various report elements into files based on what they do. So instead of having `conf.py` that includes implementation to inspect various configuration files, I propose to split files by the tool/application, which is being inspected. E.g. `tmpfiles.d`, `authselect`, `chrony`, etc... This will make it obvious what is supported and what is not. I find the current split unintuitive and hard to follow (or search what is where).
- I find the name "target" for the format of the image unfortunate, since we use this already for upload targets in composer. I propose to use something like `ImageKind` or `ImageFormat` instead.
- CI contains warnings:
    ```
    + sudo python3 setup.py install
    running install
    /usr/lib/python3.10/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
      warnings.warn(
    ```

lavocatt:
I'm closing this as the refactor will be redone under `osbuild`

PR #59, opened 2022-10-11
tools/update_tool: diff to a specific ref

lavocatt:
Before the update tool was only able to diff to the previous commit. It revealed to be unpractical for some use cases where I needed to diff between two update entries.

This commit brings this functionality.

### DIFF
--- a/tools/update_tool
+++ b/tools/update_tool
@@ -111,7 +111,7 @@ def compare(obj1, obj2):
     return "unchanged"
 
 
-def list_changed_files():
+def list_changed_files(sha1="HEAD", sha2="HEAD~1", db_dir="manifest-db"):
     """
     Go through the diff between this commit and the last one and list all the
     files in there. An update is only valid if it contains only the DB part that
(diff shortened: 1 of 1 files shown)


tools/update_tool
```diff
@@ -367,6 +368,14 @@ def main():
         default=False,
         help="Disable UUID masking"
     )
+    parser.add_argument(
+        "--compare-to",
+        default="HEAD~1",
+        help='git ref to compare to, default to previous commit')
+    parser.add_argument(
```
ondrejbudai, line 375: Am I blind, or is this unused?

lavocatt: It's used on lines 259, 294, 300 and 340 ^^.
It's just that args is directly passed to the other functions. Could be better though!

lavocatt: It's me who is blind sorry! Now the `db_path` is used :sweat_smile:

ondrejbudai (approved):
Thanks!

PR #62, opened 2022-10-14
ci: Add RHEL 8.8 and 9.2 runners

jrusz:
I have no idea if this is going to work, but this is what I figured out we need to get new manifests and image-info for enabling testing on 8.8 and 9.2 in osbuild repo.

I'm guessing this needs composer support for 8.8 and 9.2 and I don't know if that exists right now.


jrusz:
Huh, ok it won't trigger the generation from a PR... @lavocatt what is the correct procedure here? Do we want to replace the runners with new ones or do we just add them and remove the old ones later? Also am I missing something?

lavocatt:
Thanks for having a look at this!

Indeed generation is only triggered via the gitlab trigger API. You need to get a token in the gitlab repo (maybe I can help by generating one). And then you need to manually trigger the pipeline:

`curl --request POST --form token=$YOUR_TOKEN  --form ref=$PR_BRANCH_NAME https://gitlab.com/api/v4/projects/36844106/trigger/pipeline`

Here, `PR_BRANCH_NAME` would be `PR_62`

Then, when the update PR is opened, we can see if it works as we want it to. If it does, we close the update PR, Link it there and merge this PR.

Schutzfile
```diff
@@ -2,7 +2,7 @@
   "global": {
     "dependencies": {
       "osbuild": {
-        "commit": "b87eaf603284681335704910636123778dd57895"
```
lavocatt: Do the new distributions needs an updated version of OSBuild ?

lavocatt:
I've triggered the pipeline here: https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/668521912

Schutzfile
jrusz: Yes I believe so because otherwise there would be no osbuild rpms avaiable for 8.8 and 9.2

jrusz:
> I've triggered the pipeline here: https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/668521912

Cool, thanks! Let's see how that goes.

Schutzfile
lavocatt: Normally this field would be updated automatically by OSBuild's bot. In this particular case though it makes sense to update it by hand. Thanks.

lavocatt:
> > I've triggered the pipeline here: https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/668521912
> 
> Cool, thanks! Let's see how that goes.

https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/jobs/3182413610 there's no test to run for rhel88. Have we image definitions for them ?

jrusz:
> > > I've triggered the pipeline here: https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/668521912
> > 
> > 
> > Cool, thanks! Let's see how that goes.
> 
> https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/jobs/3182413610 there's no test to run for rhel88. Have we image definitions for them ?

Oh ok, does this mean it needs `osbuild-composer` support first for this? If that's the case we don't have that yet...

lavocatt:
> Oh ok, does this mean it needs `osbuild-composer` support first for this? If that's the case we don't have that yet...

Yes, we need to have the images defined in `composer`. Let's wait until they are!

lavocatt:
Note also that we can simply rerun the trigger when they are without changing anything in the code base. Since the pipeline will use the latest commit on top of `osbuild-composer` anyway. The composer reference in the `deploy.sh` is only there to install `osbuild` and its dependencies correctly.

jrusz:
@lavocatt Hi, we have 8.8 and 9.2 support in composer, can you help with updating the test cases here?

jrusz:
I've retried the jobs in the old pipeline, will that work?

lavocatt:
I think it should work now! If not I'll re-trigger the pipeline.

Schutzfile
```diff
@@ -314,6 +355,47 @@
...
+            "title": "RHEL-9-RPMREPO-NIGHTLY-CRB",
+            "name": "crb",
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el9/el9-x86_64-crb-n9.2-20221001"
+          }
+        ],
+        "aarch64": [
+          {
+            "title": "RHEL-9-RPMREPO-NIGHTLY-BaseOS",
+            "name": "baseos",
+            "baseurl": "https://rpmrepo.osbuild.org/v2/mirror/rhvpn/el9/el9-aarch64-baseos-n9.2-20221001"
```
atodorov: the repo snapshot looks rather old at this point. Maybe upgrade it?

atodorov (approved):

jrusz:
Hm, it's failing on the new runners, are we missing anything? Also, this looks weird, failures in a passing run https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/jobs/3284380631#L2921

lavocatt:
```
rhel_92-aarch64-qcow2_customize-boot Building only: FAIL: image-info failed: Traceback (most recent call last):
  File "/home/ec2-user/~/builds/redhat/services/products/image-builder/ci/manifest-db/tools/image-info", line 2742, in <module>
    main()
  File "/home/ec2-user/~/builds/redhat/services/products/image-builder/ci/manifest-db/tools/image-info", line 2736, in main
    report = analyse_image(target)
  File "/home/ec2-user/~/builds/redhat/services/products/image-builder/ci/manifest-db/tools/image-info", line 2596, in analyse_image
    append_partitions(report, device, loctl)
  File "/home/ec2-user/~/builds/redhat/services/products/image-builder/ci/manifest-db/tools/image-info", line 2528, in append_partitions
    lvm = cm.enter_context(discover_lvm(dev))
  File "/usr/lib64/python3.9/contextlib.py", line 448, in enter_context
    result = _cm_type.__enter__(cm)
  File "/usr/lib64/python3.9/contextlib.py", line 119, in __enter__
    return next(self.gen)
  File "/home/ec2-user/~/builds/redhat/services/products/image-builder/ci/manifest-db/tools/image-info", line 2461, in discover_lvm
    raise RuntimeError(r.stderr.strip())
RuntimeError: b'WARNING: Not using device /dev/loop0p3 for PV DOy3PF-98uc-BmEV-CFKf-Sw01-wTSe-BjdgCn.\n  WARNING: PV DOy3PF-98uc-BmEV-CFKf-Sw01-wTSe-BjdgCn prefers device /dev/loop1 because device was seen first.\n  Cannot activate LVs in VG rootvg while PVs appear on duplicate devices.\n  Cannot activate LVs in VG rootvg while PVs appear on duplicate devices.'
```

Image-info can't mount the LVM volumes from the built image. I'll have a look why.

For the errors you're pointing out, some manifests can't be built, usually it's OStree stuff that needs a previous commit in order to be built. Since we don't have access to a previous commit, the manifest fails validation.

lavocatt:
Yeah, the all partition mounting is suffering for rhel92 https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/jobs/3284380637

jrusz:
@lavocatt So all but 9.2 UEFI qcow boot test passed from my subset of testing https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/689988281 I had to use older snapshot for 9.2 cause otherwise there were issues with rpmbuild. So I'd say there is something wrong with the generation here or we're missing something like regeneration of test cases in `osbuild-composer` and such, I'm not familiar with how it's done now... Can you take a look at it please and try to figure out what's wrong?

lavocatt:
please don't merge yet

lavocatt (changes requested):
On hold for the moment. I'm working to fix it.

bcl:
> On hold for the moment. I'm working to fix it.

FWIW you can convert it to / from a draft to make sure it isn't accidentally merged

lavocatt:
So, I **finally** got working uploading to S3. It turns out that `image-info` does not have the same behavior on my local `fedora 36` than on `rhel92`. That means that the first one I tested did produce a result on my machine instead of an exception.
I think I will have to spin up a Rhel92 machine to continue investigating.

lavocatt:
After investigation, it turns out we've got a race condition at the `udev` level going on.

What's happenging:
```
            dev = cm.enter_context(loop_open(loctl, image, offset=start, size=size))
            read_partition(dev, part)
```
Basically, this is `create-loop` then `blkid` on the mounted patition.
Sometimes, not all the time, `blkid` can't read from the partition correctly and does not return all the information it needs.
putting a `time.sleep(3)` after `create-loop` "solves" the issue.

@gicmo told me that in OSBuild, some of the behavior defined in Udev for loopback devices are deactivated with a custom code that is not present on `image-info`:
https://github.com/osbuild/osbuild/blob/48a4419705c6427dbbf9c4ba46abb7d240832bd4/devices/org.osbuild.loopback#L69
https://github.com/osbuild/osbuild/blob/7d8b2302c0baba75b8355e27cf9229ae878857b7/osbuild/util/udev.py

One option would be to try to backport the code from https://github.com/osbuild/osbuild/blob/48a4419705c6427dbbf9c4ba46abb7d240832bd4/devices/org.osbuild.loopback to `image-info`. Another,and better option, would be to refactor `image-info` to integrate it in `osbuild`.

lavocatt:
So, just placing a `time.sleep(3)` does a bit improve things but is completely unreliable ofc. I'll spend then some time back-porting the `osbuild`'s way-to-mount to `image-info`

gicmo:
> So, just placing a `time.sleep(3)` does a bit improve things but is completely unreliable ofc. I'll spend then some time back-porting the `osbuild`'s way-to-mount to `image-info`

I think we could add a dependency on `osbuild` and then use the `DeviceManager` and `MountManager` to actually create the loopback devices and mounts. You can see how the it is being used in osbuild: https://github.com/osbuild/osbuild/blob/main/osbuild/pipeline.py#L211

To, e.g., create a loopback device, you have to instantiate a [`Device`](https://github.com/osbuild/osbuild/blob/48a4419705c6427dbbf9c4ba46abb7d240832bd4/osbuild/devices.py#L24) with the appropriate options as shown in the corresponding [`schema`](https://github.com/osbuild/osbuild/blob/48a4419705c6427dbbf9c4ba46abb7d240832bd4/devices/org.osbuild.loopback#L31).

Also see the [device tests](https://github.com/osbuild/osbuild/blob/main/test/run/test_devices.py).

lavocatt:
Closing this PR we can instead continue on this one https://github.com/osbuild/manifest-db/pull/72

lavocatt:
Waiting for https://github.com/osbuild/manifest-db/pull/72 to be merged.

lavocatt:
this can be merged after https://github.com/osbuild/manifest-db/pull/77

lavocatt:
This update was produced with the code of this PR: https://github.com/osbuild/manifest-db/pull/80 (meaning it'll work fine)

lavocatt (approved):
good for me to merge this.

PR #67, opened 2022-11-08
ci: add a second behavior for the trigger pipeline

lavocatt:
The rewrite of the image definitions makes changes to the manifests but it shouldn't affect image info.
Every change there is non-functional. It would be great if we could trigger an image-info rebuild in such cases from the PR, before it's merged, to make sure nothing has changed.

To enable this kind of testing, this pull request adds another trigger pipeline to the suite.
The desired behavior is controlled by the `TRIGGER_KIND` variable.

With its value to `default` the trigger pipeline will perform the update scenario where it will build the manifest from composer's main branch and generate the image info in order to create a pull request.

With its value to `test` the trigger pipeline will run the checks from the manifests found in the branch given as a parameter and will:

- generate the manifests from composer's main
- generate the manifests from $`COMPOSER_REMOTE`'s $ `COMPOSER_BRANCH`
- only keep the ones that changed from the remote
- build the images
- test the image-info regarding the DB

To use it:

curl --request POST\
    --form token=$YOUR_TOKEN\
    --form ref=main\
    --form "variables[COMPOSER_REMOTE]=$SOME_REMOTE" \
    --form "variables[COMPOSER_BRANCH]=$SOME_BRANCH" \
    --form "variables[TRIGGER_KIND]=test" \
    https://gitlab.com/api/v4/projects/36844106/trigger/pipeline

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -84,7 +84,45 @@ Image-info-build:
     - tools/import-image-tests manifests manifest-db --manifest-only --db-ignore=db-ignore --filter-with-ci-distros=.gitlab-ci.yml --verbose
     - sudo test/cases/manifest_tests
   rules:
-    - if: '$CI_PIPELINE_SOURCE == "trigger"'
+    - if: $CI_PIPELINE_SOURCE == "trigger" && $TRIGGER_KIND == "default"
+  artifacts:
+    when: always
+    paths:
+      - generated-image-infos/
+  dependencies:
+    - Manifest-gen
+  parallel:
+    matrix:
+      - RUNNER:
+          - aws/fedora-35-x86_64
+          - aws/fedora-35-aarch64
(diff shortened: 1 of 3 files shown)


ondrejbudai:
What's the motivation behind this PR?

achilleas-k:
> What's the motivation behind this PR?

Short answer: me :)

Longer answer: https://github.com/osbuild/osbuild-composer/pull/3120#issuecomment-1306003106
The rewrite of the definitions makes changes to the manifests but it shouldn't affect image info.  Every change there is non-functional.  It would be great if we could trigger an image-info rebuild in such cases from the PR, before it's merged, to make sure nothing has changed.

ondrejbudai:
Thanks for the explanation, it's super-useful to have some context. :) Something that should always be in the PR description.

PR #72, opened 2022-12-06
tools/image-info: use osbuild's internal to mount

lavocatt:
**Fix image-info:**
   
    On the newest versions of rhel 92, 88, a change in behavior makes the
    previous version of image-info failing to mount loopback devices. We've
    tracked down this error to be a race condition on udev, yet without
    understanding what changed for now.
    
    Osbuild had for some time already a cleaner way to mount partitions.
    osbuild has some machinery to opt out of block device handling in udev
    https://github.com/osbuild/osbuild/blob/48a4419705c6427dbbf9c4ba46abb7d240832bd4/devices/org.osbuild.loopback#L69
    Using this fixes the issue at hand.
    

  Schutzfile, Osbuild and Composer are updated.

### DIFF
--- a/tools/image-info
+++ b/tools/image-info
@@ -21,7 +21,7 @@ import xml.etree.ElementTree
 import yaml
 
 from collections import OrderedDict
-from typing import Dict
+from typing import Dict, Any, Generator
 
 from osbuild import loop
 
@@ -70,7 +70,7 @@ def loop_open(ctl, image, *, offset=None, size=None):
 
 
 @contextlib.contextmanager
-def convert_image(ctl, image, fmt):
+def convert_image(image, fmt):
     with tempfile.TemporaryDirectory(dir="/var/tmp") as tmp:
         if fmt["type"] != "raw":
             target = os.path.join(tmp, "image.raw")
@@ -135,6 +135,8 @@ def parse_environment_vars(s):
 def parse_unit_files(s, expected_state):
     r = []
     for line in s.split("\n")[1:]:
+        state = ""
+        unit = ""
         try:
             unit, state, *_ = line.split()
         except ValueError:
@@ -146,7 +148,7 @@ def parse_unit_files(s, expected_state):
     return r
 
 
-def subprocess_check_output(argv, parse_fn=None):
+def subprocess_check_output(argv, parse_fn=None) -> Any:
     try:
         output = subprocess.check_output(argv, encoding="utf-8")
     except subprocess.CalledProcessError as e:
@@ -195,7 +197,7 @@ def read_container_images(tree):
     return images
 
 
-def read_image_format(device):
+def read_image_format(device) -> Dict[str, str]:
     """
     Read image format.
 
@@ -318,7 +320,7 @@ def read_partition_table(device):
     return info
 
 
-def read_bootloader_type(device):
+def read_bootloader_type(device) -> str:
     """
     Read bootloader type from the provided device.
 
@@ -416,16 +418,18 @@ def rpm_verify(tree):
 
     changed = {}
     missing = []
-    for line in rpm.stdout:
-        # format description in rpm(8), under `--verify`
-        attrs = line[:9]
-        if attrs == "missing  ":
-            missing.append(line[12:].rstrip())
-        else:
-            changed[line[13:].rstrip()] = attrs
 
-    # ignore return value, because it returns non-zero when it found changes
-    rpm.wait()
+    if rpm.stdout:
+        for line in rpm.stdout:
+            # format description in rpm(8), under `--verify`
+            attrs = line[:9]
+            if attrs == "missing  ":
+                missing.append(line[12:].rstrip())
+            else:
+                changed[line[13:].rstrip()] = attrs
+
+        # ignore return value, because it returns non-zero when it found changes
+        rpm.wait()
 
     return {
         "missing": sorted(missing),
@@ -433,7 +437,7 @@ def rpm_verify(tree):
     }
 
 
-def rpm_not_installed_docs(tree, is_ostree):
+def rpm_not_installed_docs(tree):
     """
     Gathers information on documentation, which is part of RPM packages,
     but was not installed.
@@ -467,7 +471,7 @@ def rpm_not_installed_docs(tree, is_ostree):
     return sorted(not_installed_docs)
 
 
-def rpm_packages(tree, is_ostree):
+def rpm_packages(tree):
     """
     Read NVRs of RPM packages installed on the system.
 
@@ -494,7 +498,7 @@ def rpm_packages(tree, is_ostree):
         cmd += ["--dbpath", "/usr/share/rpm"]
     elif os.path.exists(os.path.join(tree, "var/lib/rpm")):
         cmd += ["--dbpath", "/var/lib/rpm"]
-    output = subprocess_check_output(cmd)
+    subprocess_check_output(cmd)
     pkgs = subprocess_check_output(cmd, str.split)
     return list(sorted(pkgs))
 
@@ -2271,11 +2275,11 @@ def read_resolv_conf(tree):
 
 def append_filesystem(report, tree, *, is_ostree=False):
     if os.path.exists(f"{tree}/etc/os-release"):
-        report["packages"] = rpm_packages(tree, is_ostree)
+        report["packages"] = rpm_packages(tree)
         if not is_ostree:
             report["rpm-verify"] = rpm_verify(tree)
 
-        not_installed_docs = rpm_not_installed_docs(tree, is_ostree)
+        not_installed_docs = rpm_not_installed_docs(tree)
         if not_installed_docs:
             report["rpm_not_installed_docs"] = not_installed_docs
 
@@ -2485,7 +2489,7 @@ def ensure_device_file(path: str, major: int, minor: int):
 
 
 @contextlib.contextmanager
-def discover_lvm(dev) -> Dict:
+def discover_lvm(dev) -> Generator[Dict[str, Any], None, None]:
     # find the volume group name for the device file
     vg_name = volume_group_for_device(dev)
 
@@ -2532,13 +2536,11 @@ def discover_lvm(dev) -> Dict:
             volumes[name] = info
             if name.startswith("root"):
                 volumes.move_to_end(name, last=False)
-        res = {
+        yield {
             "lvm": True,
             "lvm.vg": vg_name,
             "lvm.volumes": volumes
         }
-
-        yield res
 
     finally:
         r = subprocess.run(["vgchange", "-an", vg_name],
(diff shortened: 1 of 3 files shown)


tools/image-info
```diff
@@ -146,7 +148,7 @@ def parse_unit_files(s, expected_state):
     return r
 
 
-def subprocess_check_output(argv, parse_fn=None):
+def subprocess_check_output(argv, parse_fn=None) -> Any:
```
achilleas-k, line 147: Can't we get anything more concrete than `Any` here?

tools/image-info
```diff
@@ -2488,161 +2476,208 @@ def ensure_device_file(path: str, major: int, minor: int):
...
-        if r.returncode != 0:
-            raise RuntimeError(r.stderr.strip())
 
 
 def partition_is_lvm(part: Dict) -> bool:
     return part["type"].upper() in ["E6D6D379-F507-44C2-A23C-238F2A3DF928", "8E"]
 
 
-def append_partitions(report, image, loctl):
+def append_partitions(report, image, devpath:str, cm: contextlib.ExitStack, mgr:host.ServiceManager):
```
achilleas-k: Needs spaces after `:` for the type annotations.

tools/image-info
```diff
@@ -2488,161 +2476,208 @@ def ensure_device_file(path: str, major: int, minor: int):
...
+                offset=start)
+        )
+        devices_map[ret["path"]] = ret["dev"]
+        read_partition(ret["path"], part)
+        if partition_is_lvm(part):
+            # Temporary hack:
+            # It seems that for some distros pvdisplay doesn't manage to open
+            # the device when we're working under with /dev in the temporary
+            # directory.
+            tst = f"/dev{ret['path'].split('dev')[1]}"
```
achilleas-k: What is this doing exactly?  Is it removing a leading segment before `/dev` in a nested path?  There's probably a cleaner way to write this.  I wonder if splitting on `dev` might be an issue if any device path has more than one `dev` in it, if that's possible.

tools/image-info
```diff
@@ -2815,14 +2815,23 @@ def main():
...
+                report = analyse_directory(target)
+            elif is_tarball(target):
+                report = analyse_tarball(target)
+            elif is_compressed(target):
+                report = analyse_compressed(target)
+            else:
+                report = analyse_image(target)
+            break
+        except Exception as e:
+            print(f"failed attempt 1 {e}", file=sys.stderr)
```
achilleas-k: What kind of failures still exist here?  Is the container issue still a problem?  Has the LVM issue been fixed, or is it just less frequent?

tools/image-info
```diff
@@ -2795,6 +2796,22 @@ def analyse_compressed(path):
...
+
+
+def analyse_iso(path):
+    with tempfile.TemporaryDirectory(dir="/var/tmp") as tmp:
+        report = None
+        subprocess.run(["mount", "-o", "loop", path, tmp], check=True)
+        try:
+            report = analyse_tarball(os.path.join(tmp, "liveimg.tar.gz"))
+        except Exception as e:
+            print(f"{e}", file=sys.stderr)
```
achilleas-k, line 2854: Nothing to change for now but a note that we might support ostree installers in the future and we would need to detect the payload type here.

tools/image-info
```diff
@@ -2760,6 +2795,16 @@ def analyse_compressed(path):
         image = os.path.join(tmpdir, files[0])
         return analyse_image(image)
 
+@contextlib.contextmanager
+def make_dev_tmpfs(tmpdir):
```
achilleas-k: Did we have something similar before?  Why do we need this?

lavocatt:
On hold waiting for https://github.com/osbuild/osbuild/pull/1229.
The new image-info will will fail until the OSBuild's PR is meged and the dependencies is updated in the Schutzfile.

achilleas-k (commented):
Looks good.  Did a few local tests and couldn't reproduce any of the old issues.
I have some very minor comments below.

ondrejbudai:
regarding this commit: ` tools/image-info: import from composer`

Do we have two copies of `image-info` in both `manifest-db` and `osbuild-composer`? If that's such a case, can we remove one of them?

thozza:
> regarding this commit: ` tools/image-info: import from composer`
> 
> Do we have two copies of `image-info` in both `manifest-db` and `osbuild-composer`? If that's such a case, can we remove one of them?

We do. I'm all for deduplicating them. Ideally I would propose to move it to `osbuild` repo and packaging it in the `osbuild-tools` sub-package, so that we can use it in composer's API tests (like we already do https://github.com/osbuild/osbuild-composer/blob/7a67e7b7e8d222c7f845e7962f701b6a0d364d0b/test/cases/api/common/s3.sh#L167), as well as in manifest-db.

tools/image-info
thozza: In addition to that, I don't see a reason to pass `ExitStack` to this function. It should use `with` statement. Everything that this function does should be IMO cleaned up on the exit from it, which won't happen now. `ExitStack` is useful when the image-info previously had to mount multiple partitions in a `for` loop by using them as context managers. Other than that, I propose to not use it anywhere if it is possible to just use the `with` statement

lavocatt: We still do loop mount each partition in a loop. I extracted the ExitStack to gain one indentation level

tools/image-info
lavocatt: I'm afraid that, since the result is dependent of the dynamically given parse function, I can't get a static type on what can be returned.

tools/image-info
lavocatt: So this is part of the shady things. I wanted to stop loopback mounting in `/dev` at some point and to bind mount under `/tmpt/somethinh/dev`. So I created a tmpfs to do this, created the devpath pointing to the tmpfs and everything was working great! Except for these two distributions centos9 and rhel91 that were giving me a headache! Because on those two, and on those two only, I could not explore the file under /tmp, like I would endup having errors with lvdisplay like `Could not find parent device`.

So I implemented this hack, that is quite ugly but, functionnal...

The question I'm wondering now is, is it useful to have this `tmpfs` after all ?
So I'll will make some tests without, without the hack and see what's what.
If it works well, I'll remove the hack and the tmpfs thing.

tools/image-info
lavocatt: So this is shady thing number 2. I need to test to execute the program without it and see if it works well or not. And if it crashes, try to understand what's going on.

tools/image-info
thozza: So it is IMO still OK to use `ExitStack` in the loop.

Other than I'm not in favor of it due to the reasons I wrote above. Passing it into a function as an argument does not make sense if the function is done with all its work once it finishes. Saving indentation level is IMO not a a good reason to use it.

tools/image-info
lavocatt: I thought when adding this that our issues might come from the fact that we were interfering with the /dev of the machine in a magical-not-understood-way. Now that I'm putting this into words like that, I'm thinking that I don't have a lot of justification for this now. So I will test to remove it,see if that changes something or not.

tools/image-info
lavocatt: I don't have strong opinions on this TBH, I'll move it

lavocatt: I will just do one adjustment at a time, because, I don't want to break that working state.

tools/image-info
lavocatt: Alrighty, the test I ran this morning without the for loop makes me believe we don't need this. I'm removing it.

tools/image-info
lavocatt: @thozza So, after better inspection, it was clearly u necessary to have these ExitStack. Tanks for pointing it out! The code is better now without them.

lavocatt:
So what do we do for this PR ? Do we merge it in manifest-db (when osbuild is changed to accept mount options) or do we discard it to include image-info on osbuld ?

lavocatt:
I think that moving image-info to osbuild would be better done in a follow up.

lavocatt:
Alright, thanks for the commit landed in `osbuild` we have now a functioning `image-info` :partying_face: I've also added schema validation. So we're sure we don't send wrong parameters to `osbuild`. It's now passing the regression tests, we're making progress!

tools/image-info
```diff
@@ -533,7 +518,8 @@ def read_services(tree, state):
         "chrony-wait.service"
     ]
     """
-    services_state = subprocess_check_output(["systemctl", f"--root={tree}", "list-unit-files"], (lambda s: parse_unit_files(s, state)))
+    services_state = subprocess_check_output(
+        ["systemctl", f"--root={tree}", "list-unit-files"], (lambda s: parse_unit_files(s, state)))
```
achilleas-k, line 533: This commit mixes functional and codestyle changes.  I'd like them separate for a cleaner history.
It was generally quite hard to read this patch and I'm not entirely sure I followed most of it.

tools/image-info
```diff
@@ -2618,26 +2641,60 @@ def append_partitions(report, image, devpath:str, mountpoint: str):
...
+            # invoke the mounter. However we can reconstruct it for the json
+            # validation step. We're only interested at validate the options so
+            # we can put a placeholder string for the other values.
+            validate(info, {
+                "source": "placeholder",
+                "target": "placeholder",
+                "type": "placeholder",
+                "name": "placeholder",
+                "options": options,
+            })
```
achilleas-k: Would it be simpler to validate `info.get_schema().get("options")` against the `options` since that's the only part we care about?

achilleas-k (commented):
What's the long term plan for image-info now?  This PR copies it from composer and then fixes the issues with it.  Are we going to keep them in sync manually?  (I hope not).

If this is to be the main location of the tool, I'd recommend removing it from composer then, but first we need to replace any usage of it in osbuild-composer with the version from here.

tools/image-info
lavocatt: Indeed this commit is badly split up. I'm gonna fix that.

lavocatt: I'm gonna split this in several ones, they won't work without each other but it's gonna be easier to read through it.

tools/image-info
```diff
@@ -2519,11 +2524,15 @@ def discover_lvm(dev:str, parent:devices.Device, devmgr:devices.DeviceManager):
...
-                    name,
-                    index.get_module_info("Device", "org.osbuild.lvm2.lv"),
-                    parent,
-                    options)
+            name,
+            info,
+            parent,
+            options)
+        if not info:
+            raise RuntimeError("can't find org.osbuild.lvm2.lv")
```
achilleas-k, line 2538: This check should be above the device creation, but I suppose it's not a big issue since the Device constructor just assigns it to the property.

achilleas-k (approved):
LGTM.  Tiny note below, but nothing we can't change later.

achilleas-k:
@thozza I'm good to merge and get this into composer's CI to get rid of that flaky test for container embedding, but I'll leave it open so you can have one last look if you want to first.

PR #77, opened 2023-02-10
tools/image-info: bad syntax on rhel8 python

lavocatt:
We can't use:

with (a() as b, c() as d):
    pass

On older versions of python.

### DIFF
--- a/tools/image-info
+++ b/tools/image-info
@@ -2569,133 +2569,133 @@ def partition_is_lvm(part: Dict) -> bool:
 
 def append_partitions(report, image):
     partitions = report["partitions"]
-    with (tempfile.TemporaryDirectory() as mountpoint,
-          host.ServiceManager(monitor=monitor.NullMonitor(1)) as mgr):
+    with tempfile.TemporaryDirectory() as mountpoint:
+        with host.ServiceManager(monitor=monitor.NullMonitor(1)) as mgr:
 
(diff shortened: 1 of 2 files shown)


tools/image-info
```diff
@@ -2569,133 +2569,133 @@ def partition_is_lvm(part: Dict) -> bool:
...
-                    key = parts[0]
-                    val = parts[1]
-
-                    # uid and gid must be integers
-                    if key == "uid" or key == "gid":
-                        val = int(val)
-
-                    options[key] = val
+    with tempfile.TemporaryDirectory() as mountpoint:
+          with host.ServiceManager(monitor=monitor.NullMonitor(1)) as mgr:
```
achilleas-k: The indentation here is wrong

achilleas-k (changes requested):

tools/image-info
lavocatt: I need to add a pep8 checker somewhere! Thanks.

lavocatt:
On this PR https://github.com/osbuild/manifest-db/pull/78 I tested that when image-info is not changed on the PR the regression test immediately return. On this PR they should run because image-info was changed.

lavocatt:
PR https://github.com/osbuild/manifest-db/pull/77 contains this commit and produced the update https://github.com/osbuild/manifest-db/pull/80 so everything's tested.

ondrejbudai (approved):

achilleas-k (approved):

PR #78, opened 2023-02-10
don't merge

lavocatt:
don't merge


lavocatt:
This worked: the regression test successfully checked that there was no update to image info and therefore returned immediately.

PR #79, opened 2023-02-10
ci: test image info against DB

lavocatt:
When image image info is changed, run it against the entire data base to make sure everything is fine.


lavocatt:
I'd like to wait up until https://github.com/osbuild/manifest-db/pull/80 is merged before merging this one.

lavocatt:
We are running the tests all the time instead, it's more robust

PR #81, opened 2023-02-16
ci: update terraform reference

lavocatt:
To stop leaking CI resources, we need to be above commit sha 89fb25d15. Update to latest reference


ondrejbudai (approved):
Thanks!

PR #82, opened 2023-02-17
diff: filter tmpfiles_d if necessary

lavocatt:
Some of the files scanned by tmpfiles_d may need to be filter out
    because their content change for each build.
    The design decision made is to store in the DataBase the list of files
    a given distribution needs to filter out.
    
    Filterdiff.py gets introduced and contains the necessary bits to perform
    smart diffing between two image info. It possesses also a main function
    and can be used to diff two files. Each file can either contain a
    raw image-info json or be a DB entry.


lavocatt:
As a follow up I think I'd like to make a python lib for manifest-db as the code that is duplicated is growing up.

tools/osbuild-image-test
```diff
@@ -12,9 +12,11 @@ import subprocess
 import sys
 import traceback
 import shutil
+import re
+import tempfile
 
 from collections import defaultdict
-from typing import Iterator, List, Optional, Dict
+from typing import Iterator, List, Optional, Dict, final
```
thozza: What is the `final` for? It does not seem to exist in the `typing` module and also it does not seem to be used anywhere in the code.

tools/osbuild-image-test
```diff
@@ -27,6 +29,50 @@ RED = "\033[31m"
 YELLOW = "\033[33m"
 
 
+def filter_fn(imi):
+    """
+    Filter specific fields in the image-info that can't be compared
+    together.
+    """
+    def tmpfiles_d(imi):
+        to_remove = ["rpm-ostree-1-autovar.conf"]
```
thozza: nitpick: It would be IMO much nicer to be able to pass this to the test case as an argument, instead of hardcoding it in the test itself. Nevertheless I do not have strong opinion on this and would find this acceptable.

tools/update_tool
```diff
@@ -184,7 +198,7 @@ def filter_fn(imi):
                         if partition["fstype"] == "iso9660":
                             partition["uuid"] = "2022-07-01-fixed-uuid"
         return imi
-    return iso(lvm2(imi))
+    return iso(lvm2(tmpfiles_d(imi)))
```
thozza: It would probably make sense to keep the "sanitization" code needed for diffing in a single place. You could for e.g. move it to a separate script that would be used for pre-processing the report before diffing. And then in any workflow, you would simply call it before diffing.

tools/osbuild-image-test
```diff
@@ -503,72 +625,37 @@ def main():
...
+    parser.add_argument(
+        '--command',
+        type=str,
+        choices=[
+            "run",
+            "diff"
+        ],
+        default="run")
+    parser.add_argument(
+            "--item1",
```
thozza: nitpick: I would suggest to call his and `item2` in a more self-explanatory way, such as `base` and `new`.

tools/osbuild-image-test
lavocatt: I have no idea actually! Maybe my editor added it by mistake. I don't remember adding it.

tools/osbuild-image-test
lavocatt: I prefer leaving this as it is. I don't want to add too much parameters. And if we need to extend this list, it's good to have it where it's executed.

thozza: In the past, we ran into different issues on different distributions, Therefore applying the same rules on each distribution is not the best idea to do. But as I've said, I'm able to live with this version.

lavocatt: IC, so, I think, we should store in the DB entries which files to ignore then.

tools/osbuild-image-test
lavocatt: Now the files to diff are positional arguments, so naming convention is fixed

tools/osbuild-image-test
lavocatt: This is now stored in the DB, and only one distribution gets it (for now)

manifest-db/centos_9-x86_64-edge_commit_rt-boot.json
```diff
@@ -8096,5 +8096,6 @@
         }
       }
     }
-  }
-}
\ No newline at end of file
+  },
+  "unwanted_tmpfiled_d":["rpm-ostree-1-autovar.conf"]
```
thozza: Nitpick: I find this name a bit misleading. I would suggest something like `skip_tmpfilesd_paths`. Also the name of the property contains a typo.

In addition to that, I wonder how do you plan to handle updating manifests from composer, since I expect this to get dropped on manifest update. I think that it would be better to provide this as an argument to the test in `.gitlab-ci.yml` for the specific distro runner instead of storing it in the manifest.

If there are good reasons to keep it in the manifest, I would propose to nest this under a new top-level object such as `"image_test_config"` or something similar. We may end up adding more and it will be nicer to have them all nested under one umbrella object.

tools/filterdiff.py
```diff
@@ -0,0 +1,150 @@
...
+import sys
+import os
+import argparse
+import tempfile
+import json
+import subprocess
+from typing import List
+
+
+def filter_fn(imi, unwanted_tmpfiled_d: List):
```
thozza: As a matter of preference, I would wrap all these filtering functions in a class as classmethods, instead of nesting them inside another function.

tools/filterdiff.py
```diff
@@ -0,0 +1,150 @@
...
+        if imi:
+            if ("image-format" in imi and "type" in imi["image-format"] and
+                    imi["image-format"]["type"] == "raw"):
+                if "partitions" in imi:
+                    for partition in imi["partitions"]:
+                        if "fstype" in partition:
+                            if partition["fstype"] == "iso9660":
+                                partition["uuid"] = "2022-07-01-fixed-uuid"
+            return imi
+    return iso(lvm2(tmpfiles_d(imi, unwanted_tmpfiled_d)))
```
thozza: nitpick: this will get hard to read once more functions accept additional arguments. Calling filtering functions in a loop could be easier to read, but I do not insist on it.

thozza (commented):
Thanks for making the skipped tmpfiles.d list of files configurable!

I added a few comments. Some general nitpicks:

- There are many typos of `tmpfiled` -> `tmpfiles` in the code.
- There is a typo (`suffisent`) in the https://github.com/osbuild/manifest-db/pull/82/commits/80c25ef740db41325405506e96f8b88bbc32bb0a commit message.

manifest-db/centos_9-x86_64-edge_commit_rt-boot.json
lavocatt: I good for the naming change :)
updating won't erase the value, only the manifest part and the image-info get changed. It's isn't stored at manifest level but a level higher in the metadata parts of the DB. I can create a section metadata to hold this kind of stuff if you'd prefer.

tools/filterdiff.py
lavocatt: Good point, I'll do that!

tools/filterdiff.py
lavocatt: Can do, since I'm changing that to a class in any case. Maybe not a loop, I'll see with what I come up to. Thanks for the review!

manifest-db/centos_9-x86_64-edge_commit_rt-boot.json
thozza: My preference is to not have it defined in the image test case (what I meant by the manifest in my message, I mean basically the whole file), but provided to the test case as an argument when comparing image-info reports.

My point is that the current way it most probably works is:

1. Generate a new image test with image-info for a new image type or a new distro.
2. Then you need to manually change the image test and add any configuration for the test case.
3. Only now the image test will work.

So there is always a potential manual step for each new image types or new distros. While if it was passed as an argument to all c9s image tests, then you would not need to do anything for new image types or distros.

Lastly, using 'metadata` for the image test case configuration is IMO misleading, since it is really a configuration for the test case, not some additional data about it.

lavocatt: I'm not a huge fan of having to set this as a parameter on the command line tbh. The reason why is that it scatters the information all around. Having it in the database makes it clearer and easier to find again. If we were to set the list on the command line in the .gitlabci yaml file, we would have to set it twice. One for MDB's ci and  one for osbuild's ci, with the implication of harder maintenance.

We already have metadata in the DB entry, like the image type or the arch. For me it makes sense to embed the `skip_tmpfilesd_paths` as it is a specificity of the entry itself. The DB being there to hold the necessary information to test image production, so containing in itself the data to filter at comparison time feels straightforward to me.

```
{
  "id": "centos_9-aarch64-ami-boot",
  "desc": {
    "distro": "centos-9",
    "arch": "aarch64",
    "image-type": "ami"
  },
  "manifest": {
  },
  "image-info": {
  }
}
```

There is still the question of the addition of new distributions that will not have this parameter set to them. And I think it's fine to add it manually when necessary. This PR also adds full testing of the DB on CI push, which means that after an update is proposed, it's gonna be tested as osbuild would be. Meaning that, we will detect early the need of adding tmpfilesd to skip to the list.

thozza: > I'm not a huge fan of having to set this as a parameter on the command line tbh. The reason why is that it scatters the information all around. Having it in the database makes it clearer and easier to find again. If we were to set the list on the command line in the .gitlabci yaml file, we would have to set it twice. One for MDB's ci and one for osbuild's ci, with the implication of harder maintenance.

OK, that's a fair point.

> We already have metadata in the DB entry, like the image type or the arch. For me it makes sense to embed the `skip_tmpfilesd_paths` as it is a specificity of the entry itself. The DB being there to hold the necessary information to test image production, so containing in itself the data to filter at comparison time feels straightforward to me.
> 
> ```
> {
>   "id": "centos_9-aarch64-ami-boot",
>   "desc": {
>     "distro": "centos-9",
>     "arch": "aarch64",
>     "image-type": "ami"
>   },
>   "manifest": {
>   },
>   "image-info": {
>   }
> }
> ```

Wow, I was not aware of the fact that the image test manifests stored in `manifest-db` reinvented their actual content, compared to what we store in osbuild-composer. E.g. I was not aware that the `compose-request` https://github.com/osbuild/osbuild-composer/blob/ed4d12fe33e2eb2d09e4cd70e3e550ff8861d75d/test/data/manifests/centos_8-aarch64-ami-boot.json#L2-L22 is completely dropped in the db entry and that some values from it are instead specified under a completely different object. Not sure what was the decision of loosing the information about the compose request that was used to generate the manifest (without it there is no easy way to generate the same manifest again from composer). This is all sufficient as is for osbuild testing, but can potentially limit the usefulness of the db entry in the future. Not an issue related to this PR, but I wanted to express my surprise.

I do not feel strongly about embedding image test configuration (I still don't think that these are metadata) in the db entry itself. I think that this is OK and makes things easier when the same  configuration needs to be used in osbuild CI as well as this repo's CI.

> There is still the question of the addition of new distributions that will not have this parameter set to them. And I think it's fine to add it manually when necessary. This PR also adds full testing of the DB on CI push, which means that after an update is proposed, it's gonna be tested as osbuild would be. Meaning that, we will detect early the need of adding tmpfilesd to skip to the list.

This all sounds good. You could also define this only in the manifest-db `.gitlab-ci.yaml`, which would result in the value being embedded in the db entry (which would be then used as is in osbuild). But that's more of an enhancement that can be added later. I'm not a big fan of manual steps in workflows that can be automated, but if you are willing to do it, I'm fine with it.

lavocatt: > Wow, I was not aware of the fact that the image test manifests stored in manifest-db reinvented their actual content, compared to what we store in osbuild-composer. E.g. I was not aware that the compose-request https://github.com/osbuild/osbuild-composer/blob/ed4d12fe33e2eb2d09e4cd70e3e550ff8861d75d/test/data/manifests/centos_8-aarch64-ami-boot.json#L2-L22 is completely dropped in the db entry and that some values from it are instead specified under a completely different object.

:smile: We would have to ask Christian what he had in mind for this decision ^^

Yeah so compose request is dropped, and we only keep the few bits necessary to run the testing. One major advantage of this is that the DB can be tailored to what we want it to be. And it's kind of easy to add new elements in it if we need to. I'm not that worried about it's upgradability.

The update workflow assumes that it exists a function to generate the manifests elsewhere. Now this function is the one from composer, e.g, the first step of the update process is to fun the gen_manifests program, once it's done, the manifests are embedded in the DB and a build is triggered. Once the build is finished, image-info data is generated and included in the DB, then a PR is opened.

> This all sounds good. You could also define this only in the manifest-db .gitlab-ci.yaml, which would result in the value being embedded in the db entry (which would be then used as is in osbuild). But that's more of an enhancement that can be added later. I'm not a big fan of manual steps in workflows that can be automated, but if you are willing to do it, I'm fine with it.

I'm not a huge fan of doing a bot's job either, I'm sure we will find a proper way to update the fields automatically as the need happen.

test/cases/gen_db
```diff
@@ -0,0 +1,33 @@
...
+        f"distribution {distro}")
+
+try:
+    subprocess.run(["tools/osbuild-image-test",
+                    f"--arch={platform.machine()}",
+                    f"--distro={distro}",
+                    "--generator-mode"],
+                    check=True)
+except CalledProcessError:
+    sys.exit(1)
```
thozza, line 34: Just a thought, would it make sense to at least log the output from the exception? This way you are loosing all data about why the process failed.

tools/filterdiff.py
```diff
@@ -0,0 +1,154 @@
...
+
+class FilterImageInfo:
+    """
+    Filter specific fields in the image-info that can't be compared
+    together.
+    """
+
+    def __init__(self, imi: Optional[Dict], to_remove: Optional[List]) -> None:
+        self.imi: Dict = imi if imi else {}
+        self.to_remove = to_remove if to_remove else []
```
thozza: The `to_remove` name is too ambiguous. What to remove? (I know the answer, but this is not obvious from the name). I suggest adding the `tmpfilesd` to the name of the variable, as well as to the name of the constructor argument.

tools/filterdiff.py
```diff
@@ -0,0 +1,154 @@
...
+        c = c["image-info"]
+        # Load the unwanted files from the DB
+        if "unwanted_tmpfiles_d" in c:
+            unwanted_tmpfiles_d = c["unwanted_tmpfiles_d"]
+    return c
+
+
+def search_unwanted_tmpfiles_d(path1, path2):
+    """
+    Returns the union of
```
thozza: "Returns the union of" of what? 🤔

tools/filterdiff.py
```diff
@@ -0,0 +1,154 @@
...
+                        default=[],
+                        help="List of unwanted files to get from tmpfiles_d")
+    parser.add_argument(
+        "--visual-diff",
+        action="store_true",
+        default=False,
+        help="set to see the diff result"
+    )
+    args = parser.parse_args()
+    #to_remove = ["rpm-ostree-1-autovar.conf"]
```
thozza: this should be probably removed?

tools/filterdiff.py
```diff
@@ -0,0 +1,154 @@
...
+def load_item(path):
+    c = None
+    with open(path, "r") as f:
+        c = json.load(f)
+    # Supports raw image-info or DB entry
+    if "image-info" in c:
+        c = c["image-info"]
+        # Load the unwanted files from the DB
+        if "unwanted_tmpfiles_d" in c:
+            unwanted_tmpfiles_d = c["unwanted_tmpfiles_d"]
```
thozza: `unwanted_tmpfiles_d` is not used here at all.

tools/filterdiff.py
```diff
@@ -0,0 +1,154 @@
...
+    return c
+
+
+def search_unwanted_tmpfiles_d(path1, path2):
```
thozza: The name of this function is confusing. I would propose something like "get_unwanted_tmpfiles_d...`.

In addition to that, I propose to simplify this in the following way:

- There is IMO no benefit of accepting paths in this function. You should instead accept a single dict loaded by `load_item()` and work directly on it. But honestly this is also an overkill.
- Also you do not need to iterate over returned lists, but could use `set()` instead.

tools/filterdiff.py
```diff
@@ -0,0 +1,154 @@
...
+        if e not in p1:
+            p1.append(e)
+    return p1
+
+
+def diff(first, second, unwanted_tmpfiles_d, visual):
+    c1 = load_item(first)
+    c2 = load_item(second)
+    if not unwanted_tmpfiles_d:
+        unwanted_tmpfiles_d = search_unwanted_tmpfiles_d(first, second)
```
thozza: I propose to replace this with:

```python
unwanted_tmpfiles_d = set(unwanted_tmpfiles_d)
unwanted_tmpfiles_d.union(
    set(c1.get("unwanted_tmpfiles_d", [])),
    set(c2.get("unwanted_tmpfiles_d", []))
)
```
The benefit is that you can then remove `search_unwanted_tmpfiles_d()` completely.

tools/osbuild-image-test
```diff
@@ -412,7 +391,72 @@ def clean_output(folder):
...
+        if test.error:
+            print(f"{RED}FAIL{RESET}: {test.error}")
+            success = False
+        else:
+            print(f"{GREEN}SUCCESS{RESET}")
+
+        # in any case, clean the output folder to avoid filling out disk space
+        clean_output(args.output)
+
+    return 1 if not success else 0
```
thozza: nitpick: you could simplify this as `return 0 if success else 1` 😇

thozza (commented):
Thanks a lot for your changes. Overall this looks great from my point of view (except for comments that I added).

In addition to that, you are not consistent in using the `.py` extension for scripts in `tools/` folder. Please pick one approach and be consistent (either use it every time or don't use it at all).

tools/filterdiff.py
lavocatt: * The function is named `search` because it will try to find `unwanted_tmpfiles_d` from the input file. The input file can be an DB entry or a JSON file containing only the image_info.
* I also considered using sets, the benefit was not really obvious. I mean the loop is only two lines.

tools/filterdiff.py
lavocatt: That works only if `c1` and `c2` are DB entries. I want the diff tool to be able to diff single image-info too.

lavocatt: I'm gonna mix your solution with mine though ;)

thozza: > That works only if `c1` and `c2` are DB entries. I want the diff tool to be able to diff single image-info too.

I don't see how is that an issue. The image-info report apparently won't contain `unwanted_tmpfiles_d` key and in such case, the empty list is returned. If it is a DB entry, it will be there and its value will be returned... 🤷

tools/osbuild-image-test
lavocatt: ahah, indeed!

tools/filterdiff.py
thozza: > * The function is named `search` because it will try to find `unwanted_tmpfiles_d` from the input file. The input file can be an DB entry or a JSON file containing only the image_info.

The bottom line is that it returns the value assigned to `unwanted_tmpfiles_d` key or an empty list. It does not return all instances of it in the dict or index of it or other similar thing that I (personally) would expect from `search`.

I do not see this as a blocker though, so whatever you end up using, I can live with it. In ideal case, this function should not be needed at all IMO.

> * I also considered using sets, the benefit was not really obvious. I mean the loop is only two lines.

The benefit is that it is less code and lower complexity of operations (https://wiki.python.org/moin/TimeComplexity). The data structure is optimized for what you are trying to achieve.

lavocatt: Yeah, I agree, good points.

tools/filterdiff.py
lavocatt: It's just that `load_item` was only returning the `image-info` component of the file. Not the entire file content. And `unwanted_tmpfiles_d` is at the same level as `image-info` in the DB entry. Now `load` returns the `image-info` and a `Set` that can be empty or not depending on what was found in the file to open.

tools/filterdiff.py
```diff
@@ -0,0 +1,133 @@
...
+    unwanted_tmpfiles_d in the file, return it also=.
+    """
+    c = None
+    with open(path, "r") as f:
+        c = json.load(f)
+    # Supports raw image-info or DB entry
+    if "image-info" in c:
+        c = c["image-info"]
+    unwanted_tmpfiles_d = set(c.get("unwanted_tmpfiles_d", []))
+    return c, unwanted_tmpfiles_d
```
thozza: Thank you, I very appreciate your rework.

At the same time I'm sorry, but I don't like the end result. My reasons are:

- You may end up having to filter more than just tmpfilesd entries in the future and in such case the `load_item()` number of return values will just explode.
- The line `c = c["image-info"]` will always override what is in the `c` with `image-info` key value if it exists, therefore there won't be any `unwanted_tmpfiles_d` key if this is a DB entry.

It seems to me that many the approach of trying to handle multiple situations at once unnecessarily complicates the logic.

Both `diff()` and `FilterImageInfo()` seem to be operating only on `image-info` report if it exists in the loaded file.

I would propose to simply return the `dict` as loaded from the file. Then in the `diff()` get the `unwanted_tmpfiles_d` from each entry + something like `imi = c.get("image-info", [])`. Run the returned list through `FilterImageInfo()` and then diff the result. If you feel bad about duplicating 3 lines, you can wrap this pre-processing in the `load()` or a different function and return the end result for diffing.

thozza (commented):
I added one comment, otherwise all looks good 👍

tools/filterdiff.py
lavocatt: There is great value at handling the two situations. The reason why:
* You got an comparison mismatch on the CI, you wan't to check what's going on, you download the image info and run the comparison with what's int he DB. You then need to be able to compare a DB entry with an image-info.
So I insist on the need to handle this two usecases.

Good catch for the unwanted override. Too much rewrites lead to meh code sometimes.

thozza: I agree and do not dispute the usefulness of handling the two situations by the code. What I'm trying to suggest is to concentrate the differences ideally in a single function and then if possible, make the rest of the code agnostic to the fact if the input was image-info report of a DB entry. Which you kind of already have. My suggestion is to basically have all the double situation handling in `diff()`. I admit that there is no ideal solution, so I'll be happy with any outcome that does not have the bug from rewrite...

lavocatt: Thanks for clarifying :smile: I'm gonna give this a bit more thoughts then.

lavocatt: So I think a lot was to be improved in the previous implementation. I gave a complete overhaul on how the ImageInfo is loaded from files for filterdiff but also improved the way the diffing was invoked in the other two executables.

tools/filterdiff.py
```diff
@@ -0,0 +1,161 @@
...
+            return ImageInfo(
+                c["image-info"],
+                set(c.get("unwanted_tmpfiles_d", []))
+            )
+        return ImageInfo(
+            c,
+            set()
+        )
+
+    def union_unwanted_tmpfiles_d(self, set_lst: List[Set[str]]) -> Set[str]:
```
thozza: nitpick: I don't think that saving the one letter `i` from `list` to make it `lst` is worth the decreased readability. You could have called it just `sets`.

tools/filterdiff.py
```diff
@@ -0,0 +1,161 @@
...
+
+    imi: Dict
+    unwanted_tmpfiles_d: Set
+
+    def __init__(self, imi: Dict, unwanted_tmpfiles_d: Set, do_filter=True) -> None:
+        self.unwanted_tmpfiles_d = unwanted_tmpfiles_d
+        self.imi = imi
+        self.do_filter = do_filter
+
+    @classmethod
```
thozza, line 80: Why `@classmethod` and not `@staticmethod`? The method does not use anything from the class.

tools/filterdiff.py
```diff
@@ -0,0 +1,161 @@
...
+    parser.add_argument(
+        "--visual-diff",
+        action="store_true",
+        default=False,
+        help="set to see the diff result"
+    )
+    args = parser.parse_args()
+
+    imi1 = ImageInfo.from_file(args.Files[0])
+    imi2 = ImageInfo.from_file(args.Files[1]),
```
thozza: Forgotten `,` at the end of the line?

tools/filterdiff.py
```diff
@@ -0,0 +1,161 @@
...
+        # Supports raw image-info or DB entry
+        if "image-info" in c:
+            return ImageInfo(
+                c["image-info"],
+                set(c.get("unwanted_tmpfiles_d", []))
+            )
+        return ImageInfo(
+            c,
+            set()
+        )
```
thozza: Judging from the `main()` function, you probably wanted to return only a single value here, specifically an instance of `ImageInfo` class from this method.

tools/filterdiff.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def union_unwanted_tmpfiles_d(self, set_lst: List[Set[str]]) -> Set[str]:
+        res: Set[str] = self.unwanted_tmpfiles_d
+        for st in set_lst:
+            res = res.union(st)
```
thozza: Just use `res = res.union(*set_lst)` instead. (https://docs.python.org/3/reference/expressions.html#expression-lists)

tools/filterdiff.py
```diff
@@ -0,0 +1,161 @@
...
+        return res
+
+    def diff(self, other, unwanted_tmpfiles_d: Set, verbose=False) -> bool:
+        if not self.imi:
+            return not other.imi
+        unwanted_tmpfiles_d = self.union_unwanted_tmpfiles_d([
+            unwanted_tmpfiles_d,
+            other.unwanted_tmpfiles_d])
```
thozza: I suggest to delete `union_unwanted_tmpfiles_d()` completely and replace this with
```python
unwanted_tmpfiles_d = unwanted_tmpfiles_d.union(
    self.unwanted_tmpfiles_d,
    other.unwanted_tmpfiles_d
)
```

tools/filterdiff.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def union_unwanted_tmpfiles_d(self, set_lst: List[Set[str]]) -> Set[str]:
+        res: Set[str] = self.unwanted_tmpfiles_d
+        for st in set_lst:
+            res = res.union(st)
+        return res
+
+    def diff(self, other, unwanted_tmpfiles_d: Set, verbose=False) -> bool:
+        if not self.imi:
+            return not other.imi
```
thozza: This is IMHO a bit hard to follow. What's the risk of not special-casing this and just compare the two image-info reports?

tools/filterdiff.py
lavocatt: I should've user `cls(..)` instead of `ImageInfo(...)` that's why I used a `classmethod` in the first place. I just forgot why in the middle of the process.

tools/filterdiff.py
lavocatt: Exactly, isn't it what's done here ?

tools/filterdiff.py
thozza: Right, makes sense (once you'll use it 😅 ).

tools/filterdiff.py
lavocatt: Totally

tools/filterdiff.py
thozza: ah, sorry. I got confused by the formatting. 🤦  Nevermind.

tools/filterdiff.py
lavocatt: Sure it'll work without it quite fine actually. Since an ImageInfo object can't be instantiated with None values.

tools/filterdiff.py
lavocatt: Yeah, "inattention" errors are my forte :sweat_smile:

tools/filterdiff.py
lavocatt: No worries!

lavocatt: I also should do better than to write confusing code in the first place.

tools/filterdiff.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(description="osbuild image tests")
+    parser.add_argument("files",
+                        nargs=2,
+                        help="files to diff, 2 required")
+    parser.add_argument("--unwanted-tmpfiles-d",
+                        nargs="+",
+                        default=[],
```
thozza, line 162: nitpick: you could have defined this option as:

```python
parser.add_argument("--unwanted-tmpfiles-d",
    nargs="+",
    default=set(),
    type=set,
    help="List of unwanted files to get from tmpfiles_d")
```

By specifying the `type`, [parser can automatically convert the user-provided value](https://docs.python.org/3/library/argparse.html#type).

tools/osbuild-image-test
```diff
@@ -148,19 +150,23 @@ class TestCase:
     can dump the built image-info to disk
     """
 
-    def __init__(self, path, data: Dict, osb: OSBuild, imi: ImageInfo) -> None:
+    built_image_info: ImageInfo
```
thozza: nitpick: there's AFAIK no benefit in defining the variable on the class level, since it is defined in the constructor. Also you are not doing it for any other variables.

thozza (approved):
Thanks for the changes. Functionally all looks good, but there are two cosmetic things that could be improved. I'm OK with a follow-up though if you feel like this PR was already too much 😅 .

tools/osbuild-image-test
lavocatt: The thing is, `built_image_info` can't be defined in the constructor. So it's type would have to be an `Optional[ImageInfo]` type. Setting the variable at class level allows me to not face the error `attribute first seen outside of init` (or something equivalent) and also allows me to avoid aving to deal with a field that can be `Optional` Because I would have to include checks. But in fairness, adding checks could be a good idea too!

lavocatt:
> Thanks for the changes. Functionally all looks good, but there are two cosmetic things that could be improved. I'm OK with a follow-up though if you feel like this PR was already too much sweat_smile .

Nah that's fine ;) Thanks for your review. I always tend to promise I'll do something in a follow up and endup forgetting about it. So i'll change that today and we can merge this on Monday.

tools/osbuild-image-test
thozza: I don't see how this could happen, since you always assign `ImageInfo()` instance to it in the constructor 🤔

lavocatt: I'm settling for initializing an empty one.

tools/filterdiff.py
lavocatt: Hmmm, I can't get that to work :thinking: 
While printing what's in `args.unwanted-tmpfiles-d ` from the command `python tools/filterdiff.py manifest-db/centos_9-s390x-qcow2-boot.json manifest-db/centos_9-s390x-qcow2_customize-boot.json --unwanted-tmpfiles-d toto --unwanted-tmpfiles-d titi`
I get:
`[{'i', 't'}]`
Which is not what I want ^^

lavocatt: However I did find another tiny bug along the way.

thozza: I see 🤔  This is the result of using `type=set`, which I didn't realize when suggesting. What the code does is basically `set(<string>)` which ends up being what you see... You could implement your own handling, but in this case it is probably for the best to just do what you did originally.

thozza (approved):
🎉  Thanks for your patience 🙇

PR #84, opened 2023-03-15
schutzfile: update osbuild ref 2023-03-15

schutzbot:
This PR updates the osbuild ref dependency for manifest-db. Between the
last time it was updated, and this new reference commit, these are the changes:

- https://github.com/osbuild/osbuild/commit/2848867 ci: update manifest tests
- https://github.com/osbuild/osbuild/commit/ceb4b94 pkg: osbuild should depend on skopeo
- https://github.com/osbuild/osbuild/commit/3634120 Post release version bump
- https://github.com/osbuild/osbuild/commit/345b2a5 stages/ignition: support multi kargs in network kcmdline
- https://github.com/osbuild/osbuild/commit/996c1ca ci: update terraform reference
- https://github.com/osbuild/osbuild/commit/b5b3a5b Post release version bump
- https://github.com/osbuild/osbuild/commit/200c2b0 stages/copy: add option to remove destination before copying
- https://github.com/osbuild/osbuild/commit/d3229dc stages/shell.init: add top-level files property
- https://github.com/osbuild/osbuild/commit/eb903a9 stages/shell.init: restrict filename pattern
- https://github.com/osbuild/osbuild/commit/b848e5d stages/shell.init: add pattern for env var names
- https://github.com/osbuild/osbuild/commit/291726a Post release version bump

PR #85, opened 2023-03-23
tools/image-info: properly read systemD drop-in

lavocatt:
ConfigParser is not able to properly load the SystemD drop-in files. Because:

* Multiple assignments to the same option in systemd units are concatenated with white spaces.
* Multiple sections with the same section name are merged.
* An option assignment of empty value resets the option.

See https://github.com/python/cpython/issues/75709

Introduce SystemDConfigParser which has the wanted behavior.

### DIFF
--- a/tools/image-info
+++ b/tools/image-info
@@ -1553,17 +1553,10 @@ def read_systemd_service_dropin(dropin_dir_path):
     """
     # read all unit drop-in configurations
     config_files = glob.glob(f"{dropin_dir_path}/*.conf")
-    parser = configparser.RawConfigParser()
-    # prevent conversion of the opion name to lowercase
-    parser.optionxform = lambda option: option
-    parser.read(config_files)
 
     dropin_config = {}
-    for section in parser.sections():
-        section_config = {}
-        section_config.update(parser[section])
-        if section_config:
-            dropin_config[section] = section_config
+    for file in config_files:
+        dropin_config[os.path.basename(file)] = read_config_file_no_comment(file)
 
     return dropin_config
 
@@ -1599,9 +1592,9 @@ def read_systemd_service_dropins(tree):
     return _read_glob_paths_with_parser(tree, checked_globs, read_systemd_service_dropin)
 
 
-def read_tmpfilesd_config(config_path):
+def read_config_file_no_comment(config_path):
     """
-    Read tmpfiles.d configuration files.
+    Read configuration files.
 
     Returns: list of strings representing uncommented lines read from the
     configuration file.
@@ -1637,7 +1630,7 @@ def read_tmpfilesd_configs(tree):
     - "/usr/lib/tmpfiles.d/*.conf"
 
     Returns: dictionary as returned by '_read_glob_paths_with_parser()' with
-    configuration representation as returned by 'read_tmpfilesd_config()'.
+    configuration representation as returned by 'read_config_file_no_comment()'.
 
     An example return value:
     {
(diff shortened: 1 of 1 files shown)


lavocatt:
Tested alongside #84 to validate that it fixes the issue with the newer OSBuild version.

achilleas-k:
Are we going with just reading the non-comment lines for this instead?

thozza:
> Are we going with just reading the non-comment lines for this instead?

+1 for just reading non-commented lines. TBH, this looks like an overkill to me for the purpose we use image-info.

My suggestion would be to e.g. rename https://github.com/osbuild/manifest-db/blob/e918b6adfe135c006f16cbf4976a8add0522883c/tools/image-info#L1602-L1627 to something generic and then use it in the case of unit dropins as well as for tmpfilesd configs.

lavocatt:
> > Are we going with just reading the non-comment lines for this instead?
> 
> +1 for just reading non-commented lines. TBH, this looks like an overkill to me for the purpose we use image-info.
> 
> My suggestion would be to e.g. rename
> 
> https://github.com/osbuild/manifest-db/blob/e918b6adfe135c006f16cbf4976a8add0522883c/tools/image-info#L1602-L1627
> to something generic and then use it in the case of unit dropins as well as for tmpfilesd configs.

Why using it for `tmpfilesd` ? I feel this is out of the scope of this PR.

thozza:
> Why using it for `tmpfilesd` ? I feel this is out of the scope of this PR.

So the function `read_tmpfilesd_config()` has literally `tmpfilesd` in its name, since it is used for reading tmpfilesd configuration files. What I'm saying is that its implementation is IMO exactly what would be sufficient also for reading systemd unit files. So in case you'll use it instead of the INI parser, I propose to rename it to something generic (like `read_uncommented_lines()`) and call it from two places (tmpfilesd config handling AND systemd unit handling).

If you want to do it fancier, you could pass characters that we used to distinguish comments as an argument and maybe use it elsewhere as well. But that is just a super-nice-to-have and out of scope IMO.

lavocatt:
> > Why using it for `tmpfilesd` ? I feel this is out of the scope of this PR.
> 
> So the function `read_tmpfilesd_config()` has literally `tmpfilesd` in its name, since it is used for reading tmpfilesd configuration files. What I'm saying is that its implementation is IMO exactly what would be sufficient also for reading systemd unit files. So in case you'll use it instead of the INI parser, I propose to rename it to something generic (like `read_uncommented_lines()`) and call it from two places (tmpfilesd config handling AND systemd unit handling).
> 
> If you want to do it fancier, you could pass characters that we used to distinguish comments as an argument and maybe use it elsewhere as well. But that is just a super-nice-to-have and out of scope IMO.

Ok, my eyes were not in front of the eyeballs. I get you now.

tools/image-info
```diff
@@ -1596,10 +1566,10 @@ def read_systemd_service_dropins(tree):
...
     ]
 
-    return _read_glob_paths_with_parser(tree, checked_globs, read_systemd_service_dropin)
+    return _read_glob_paths_with_parser(tree, checked_globs, read_config_file_no_comment)
 
 
-def read_tmpfilesd_config(config_path):
+def read_config_file_no_comment(config_path):
     """
     Read tmpfiles.d configuration files.
```
thozza: Nitpick, it would be good to change the doc text of the function.

thozza (commented):
I think that the doc text of the renamed function should be amended. The rest looks good 👍

thozza (approved):

PR #88, opened 2023-04-18
ci: improve parallelisation

lavocatt:
Run more than once the same arch/distro in parallel. Each instance will work on the other half of the test cases to run. All tests cases will be executed but twice as fast.

### DIFF
--- a/test/cases/gen_db
+++ b/test/cases/gen_db
@@ -21,14 +21,23 @@ distro = distro.replace('"', '')
 distro = distro.replace('.', '')
 
 print(f"Running the osbuild-image-test for arch {platform.machine()} and "
-        f"distribution {distro}")
+      f"distribution {distro}")
 
+command = ["tools/osbuild-image-test",
+           f"--arch={platform.machine()}",
+           f"--distro={distro}",
+           "--generator-mode"]
+
+pa_exec = sys.argv[1] if len(sys.argv) > 1 else None
+if pa_exec:
+    total = int(pa_exec.split("/")[1])
+    part = int(pa_exec.split("/")[0])
+    command.append(f"--instance-number={part}")
+    command.append(f"--total-number-of-instances={total}")
+
+print(" ".join(command), flush=True)
 try:
-    subprocess.run(["tools/osbuild-image-test",
-                    f"--arch={platform.machine()}",
-                    f"--distro={distro}",
-                    "--generator-mode"],
-                    check=True)
+    subprocess.run(command, check=True)
 except CalledProcessError as e:
     print(e, file=sys.stderr)
     sys.exit(1)
--- a/test/cases/manifest_tests
+++ b/test/cases/manifest_tests
@@ -20,13 +20,23 @@ distro = f"{config.get('DEFAULT', 'ID')}-{config.get('DEFAULT', 'VERSION_ID')}"
 distro = distro.replace('"', '')
 distro = distro.replace('.', '')
 
-print(f"Running the osbuild-image-test for arch {platform.machine()} and "
-        f"distribution {distro}")
 
+print(f"Running the osbuild-image-test executable for arch {platform.machine()} and "
+      f"distribution {distro}")
+
+command = ["tools/osbuild-image-test",
+           f"--arch={platform.machine()}",
+           f"--distro={distro}"]
+
+pa_exec = sys.argv[1] if len(sys.argv) > 1 else None
+if pa_exec:
+    total = int(pa_exec.split("/")[1])
+    part = int(pa_exec.split("/")[0])
+    command.append(f"--instance-number={part}")
+    command.append(f"--total-number-of-instances={total}")
+
+print(" ".join(command), flush=True)
 try:
-    subprocess.run(["tools/osbuild-image-test",
-                    f"--arch={platform.machine()}",
-                    f"--distro={distro}"],
-                    check=True)
+    subprocess.run(command, check=True)
 except CalledProcessError:
     sys.exit(1)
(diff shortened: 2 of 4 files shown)


tools/osbuild-image-test
```diff
@@ -402,7 +402,21 @@ def run_tests_cases(args):
...
+    #
+    # If the user specified that more than one instances are sharing the tests
+    # set, then figure out which tests to execute based on the instance number
+    # over the total number of instances.
+    bucket_number: int = args.instance_number - 1
+    bucket_size: int = math.ceil(len(tests) / args.nb_instances)
+
+    start_index: int = bucket_number * bucket_size
+    end_index: int = (bucket_number + 1) * bucket_size
+    end_index = end_index if end_index <= len(tests) else len(tests)
```
ondrejbudai:
```suggestion
    end_index = min(end_index, len(tests))
```

tools/osbuild-image-test
```diff
@@ -402,7 +402,21 @@ def run_tests_cases(args):
...
+
+    for test in tests[start_index: end_index]:
```
ondrejbudai: There is usually no whitespace inside the Python's slicing operator:

```suggestion
    for test in tests[start_index:end_index]:
```

test/cases/gen_db
```diff
@@ -21,14 +21,23 @@ distro = distro.replace('"', '')
...
+           "--generator-mode"]
+
+pa_exec = sys.argv[1] if len(sys.argv) > 1 else None
+if pa_exec:
+    total = int(pa_exec.split("/")[1])
+    part = int(pa_exec.split("/")[0])
+    command.append(f"--instance-number={part}")
+    command.append(f"--nb-instances={total}")
+
+print(" ".join(command), flush=True)
```
ondrejbudai, line 38: A stray print? Or do you want to leave this in?

test/cases/manifest_tests
```diff
@@ -20,13 +20,23 @@ distro = f"{config.get('DEFAULT', 'ID')}-{config.get('DEFAULT', 'VERSION_ID')}"
...
+           f"--distro={distro}"]
+
+pa_exec = sys.argv[1] if len(sys.argv) > 1 else None
+if pa_exec:
+    total = int(pa_exec.split("/")[1])
+    part = int(pa_exec.split("/")[0])
+    command.append(f"--instance-number={part}")
+    command.append(f"--nb-instances={total}")
+
+print(" ".join(command), flush=True)
```
ondrejbudai, line 38: Same here.

tools/osbuild-image-test
```diff
@@ -540,6 +554,18 @@ def main():
         default="manifest-db",
         help="path to the database folder"
     )
+    parser.add_argument(
+        "--instance-number",
+        type=int,
+        default=1,
+        help="if several identical instances are running specify it's number here"
```
ondrejbudai:
```suggestion
        help="if several identical instances are running, specify its number here"
```

tools/osbuild-image-test
```diff
@@ -540,6 +554,18 @@ def main():
...
+    )
+    parser.add_argument(
+        "--nb-instances",
```
ondrejbudai: I wonder: What does `nb` mean? :)

ondrejbudai (changes requested):
Simple, yet very effective, I like it! Some comments inline, nothing major, I believe we should be able to merge this soon. :)

tools/osbuild-image-test
lavocatt: Ahah, that might be a French habit ? lol, It's for the total number of instances, maybe renaming it `--total-number-of-instances` might be better ? What do you prefer ?

test/cases/gen_db
lavocatt: I'd rather leave that in, a bit of debug, might be always useful. You know just to know how many instances are running. Also I think, I'm gonna add a bit more logs in the main program, to make sure we always run all the test cases.

tools/osbuild-image-test
lavocatt: This Autopep8 is great, except when it doesn't run on non .py files that python ones. Good catch thanks!!

tools/osbuild-image-test
lavocatt: This is better!

lavocatt:
> Simple, yet very effective, I like it! Some comments inline, nothing major, I believe we should be able to merge this soon. :)

Thanks!! I'm gonna make another round of improvement in the parallelization. Some instances are taking 2 hours to complete, I'm gonna target these ones to split them in 4 pieces instead of 2 to speed up everything.

tools/osbuild-image-test
lavocatt: Interesting, after running Autopep8 on the file, it doesn't care about this whitespace. I'm removing it manually.

lavocatt:
Before we can accept this PR, I will need to run a fake update against it to see if the workflow is still valid with parallelization.

tools/osbuild-image-test
ondrejbudai: `--num-instances` is what I would use in English. :)

ondrejbudai (approved):

tools/osbuild-image-test
lavocatt: Do you want me to change it ? Or is  `--total-number-of-instances` good enough ?

ondrejbudai: Well, it's quite long, but also descriptive. I don't mind either, feel to merge, or rename, I have no strong preference. :)

PR #91, opened 2023-05-03
Update terraform SHA

jrusz:
Centos-stream-9 images are gone for some reason, let's use new ones.


lavocatt (approved):
LGTM! And many thanks for fixing this!

PR #98, opened 2023-11-20
Return 1 when no test cases are found

jrusz:


tools/osbuild-image-test
```diff
@@ -400,7 +400,7 @@ def run_tests_cases(args):
     success = True
     if not tests:
         print(f"{RED}FAIL{RESET}: no tests to run")
-        return False
+        return 0
```
atodorov: This is also the return code for `main()` which is also the exit code for the script, so it should be non-zero!

atodorov (changes requested):

tools/osbuild-image-test
jrusz: :facepalm: Omg, yes, I wanted to put `return 1` actually, thanks!

atodorov (approved):

tools/osbuild-image-test
```diff
@@ -400,7 +400,7 @@ def run_tests_cases(args):
     success = True
     if not tests:
         print(f"{RED}FAIL{RESET}: no tests to run")
-        return False
+        return 1
```
mvo5, line 403: I am very new to this code so please pardon me if I make silly suggestions. I wonder if instead the function should be more "pythonic", i.e. return True/False and then that is converted only on the call site to the return code, something like the following diff maybe? this also calls the variable retcode (or maybe exit_code?) to make it clear that it's not a boolean -  wdyt?
```python
diff --git a/tools/osbuild-image-test b/tools/osbuild-image-test
index 4f3dd8d..2a12922 100755
--- a/tools/osbuild-image-test
+++ b/tools/osbuild-image-test
@@ -461,7 +461,7 @@ def run_tests_cases(args):
         # in any case, clean the output folder to avoid filling out disk space
         clean_output(args.output)
 
-    return 0 if success else 1
+    return success
 
 
 def main():
@@ -569,8 +569,11 @@ def main():
     )
 
     args = parser.parse_args()
-    return run_tests_cases(args)
-
+    if run_tests_cases(args):
+        retcode = 0
+    else:
+        retcode = 1
+    return retcode
 
 if __name__ == "__main__":
     sys.exit(main())
```
or as a variant
```python
diff --git a/tools/osbuild-image-test b/tools/osbuild-image-test
index 4f3dd8d..98ea703 100755
--- a/tools/osbuild-image-test
+++ b/tools/osbuild-image-test
@@ -461,7 +461,7 @@ def run_tests_cases(args):
         # in any case, clean the output folder to avoid filling out disk space
         clean_output(args.output)
 
-    return 0 if success else 1
+    return success
 
 
 def main():
@@ -569,8 +569,9 @@ def main():
     )
 
     args = parser.parse_args()
-    return run_tests_cases(args)
-
+    success = run_tests_cases(args):
+    retcode 0 if success else 1
+    return retcode
 
 if __name__ == "__main__":
     sys.exit(main())
```

mvo5 (commented):
Hey, just a quick drive-by comment as I saw this flying by. Hope it's useful :)

tools/osbuild-image-test
jrusz: Sure why not, looks better, feel free to commit this :) .. I'm mainly concerned with the fact, that the CI job passes despite logs showing failure because no tests were collected to be executed.

mvo5: Thanks! I will do followup then :) Totally fine to merge as is as it's ready

jrusz: Yeah, I just need to figure out how to add the missing test cases :sweat_smile:

lavocatt: This software is going to be decommissioned soon, you're very welcome to hack it as much as you want, just know that the result is not going to live long :slightly_smiling_face:

jrusz: I suppose this change is not needed after all, marking as resolved.

PR #100, opened 2023-12-08
Add additional ignore entries

jrusz:
This repo is no longer being maintained. In order to get at least some testing for new nightlies I'm adding all currently failing entries to the db-ignore file. Most of them are ostree/edge which should not work here in the first place.


jrusz:
Considering this repo is basically in deprecation mode we agreed with @lavocatt that makes sense to just add all the currently failing test cases to the `db-ignore` to get at least some testing on the new nightlies before this can be taken over by the images repository. Most of them are `edge` type images which should be there in the first place, the only one that really stood out to me is `ova` image type. I just hope it won't create a false sense of "all is good" but since we're testing all the important image types in integration tests in osbuild-composer repository I think we should be good. The pipeline passed here https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/pipelines/1100165687 when I tried it manually before opening this MR and Schutzbot even created a PR to update the db: https://github.com/osbuild/manifest-db/pull/99

What do you think about this @atodorov @achilleas-k ?

achilleas-k:
I think that's fine.  I have some thoughts about changing this repo to download images and manifests from the ones that are generated in osbuild/images CI and maintain image-info on those, as a reference database.  But no one has time to work on that now so let's keep this as it is for now, make it green, and revisit it another time.

achilleas-k (approved):

PR #101, opened 2023-12-14
tools/image-info: Update Mount() signature.

dustymabe:
This is needed for when https://github.com/osbuild/osbuild/pull/1501 merges. That PR adds the ability to mount partition device from a defined source rather than the source itself directly and the partition option to the function was added to allow requesting this behavior.

### DIFF
--- a/tools/image-info
+++ b/tools/image-info
@@ -2678,13 +2678,20 @@ def append_partitions(report, image):
                 jsonschema.validate(options, info.get_schema()["properties"]["options"])
 
                 # Finally mount
-                mmgr.mount(mounts.Mount(
-                    part_device,
-                    info,
-                    devices_map[part_device],  # retrieves the associated Device Object
-                    part_mountpoint,
-                    options))
-
+                mnt_kwargs = {
+                    "name": part_device,
+                    "info": info,
+                    # retrieves the associated Device Object
+                    "device": devices_map[part_device],
+                    "target": part_mountpoint,
+                    "options": options,
+                }
+                # XXX: remove inspect and just add once osbuild PR#1501 is merged
+                import inspect
+                if "partition" in inspect.getfullargspec(mounts.Mount):
+                    # Just a filesystem, no partitions on this device
+                    mnt_kwargs["partition"] = None
+                mmgr.mount(mounts.Mount(**mnt_kwargs))
             if not root_tree:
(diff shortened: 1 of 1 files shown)


dustymabe:
CI failed here I think because it needs https://github.com/osbuild/osbuild/pull/1501 to merge first. Unfortunately we're going to have to merge one of this PR or that one without full CI passing I think because the changes are coupled together.

tools/image-info
```diff
@@ -2682,6 +2682,7 @@ def append_partitions(report, image):
                     part_device,
                     info,
                     devices_map[part_device],  # retrieves the associated Device Object
+                    None,  # Just a filesystem, no partitions on this device
```
mvo5: This looks fine, if you want it to work without a force merge this probably helps (untested but should work):
```python
diff --git a/tools/image-info b/tools/image-info
index da2f8d3..d2feafd 100755
--- a/tools/image-info
+++ b/tools/image-info
@@ -2678,13 +2678,20 @@ def append_partitions(report, image):
                 jsonschema.validate(options, info.get_schema()["properties"]["options"])
 
                 # Finally mount
-                mmgr.mount(mounts.Mount(
-                    part_device,
-                    info,
-                    devices_map[part_device],  # retrieves the associated Device Object
-                    part_mountpoint,
-                    options))
-
+                mnt_kwargs = {
+                    "name": part_device,
+                    "info": info,
+                    # retrieves the associated Device Object
+                    "device": devices_map[part_device],
+                    "target": part_mountpoint,
+                    "options": options,
+                }
+                # XXX: remove inspect and just add once osbuild PR#1501 is merged
+                import inspect
+                if "partition" in inspect.getfullargspec(mounts.Mount):
+                    # Just a filesystem, no partitions on this device
+                    mnt_kwargs["partition"] = None
+                mmgr.mount(mounts.Mount(**mnt_kwargs))
             if not root_tree:
                 raise RuntimeError("The root filesystem tree is not mounted")
 
```

mvo5 (approved):
Thank you, one suggestion inline how to avoid the CI failure (a bit heavy handed maybe but we can just remove it after and it's a nice feeling to have green CI :)

tools/image-info
dustymabe: Updated the code to match your suggestion!

mvo5 (approved):
Looks fine, thank you!

PR #106, opened 2024-01-02
Fix updating of the manifest DB

thozza:
https://github.com/osbuild/manifest-db/pull/103 missed the fact that the `cmd/gen-manifests` from `osbuild/images` depends on `dnf-json` installed by `osbuild-depsolve-dnf` package. Fix this now.

Example failure: https://gitlab.com/redhat/services/products/image-builder/ci/manifest-db/-/jobs/5841293230


lavocatt (dismissed):
Excellent! Thanks for the fix

achilleas-k (dismissed):
Wonderful!  Thank you.

jrusz (commented):
It's failing to install on centos-stream and rhel-nightly both 8 and 9, only GA rhel are working. It seems DNF can't find new enough `osbuild-selinux` package. I would expect it to be included the osbuild repo though...

EDIT: Actually I guess it needs `selinux-policy` and `selinux-policy-targeted` in newer version, so I guess newer rpmrepo snapshots?

jrusz:
`osbuild` repo has snaphots from `20231201`, can you try updating them here as well @thozza ?

thozza:
> `osbuild` repo has snaphots from `20231201`, can you try updating them here as well @thozza ?

yeah, I just found out about this dependency hell :face_with_spiral_eyes:

thozza:
I resolved the dependency issue by updating used snapshots and also cleaned up a few other things... The repo could use some more love, such as not using EOL Fedora releases, but this is out of scope for my PR...

jrusz (approved):
LGTM

PR #115, opened 2024-02-20
Fix manifest parsing

tkoscieln:


ochosi:
Could you please add some context to the PR description on the specific problem/bug this PR is trying to solve? (Ideally including instructions on how to reproduce the problem)

tkoscieln:
> Could you please add some context to the PR description on the specific problem/bug this PR is trying to solve? (Ideally including instructions on how to reproduce the problem)

This was part of an effort to generate new manifests for RHEL 9.3/8.9 GA in order to migrate manifest tests in osbuild. With the change in notation where dot separator was added between the major and minor version, the manifest generation broke as the old notation without the separator was hardcoded in multiple places. I closed this as it was already merged in the automated db update that was run from this PR.

PR #134, opened 2024-07-19
Introduce pre commit

schuellerf:
Trigger was #133 where `Schutzfile` had a typo and it took a while to find it

### DIFF
--- a/.github/workflows/auto_update.yml
+++ b/.github/workflows/auto_update.yml
@@ -1,9 +1,9 @@
 name: Update DB
 
-on:
+on:  # yamllint disable-line rule:truthy
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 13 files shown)


schuellerf:
To test, this should be sufficient:
```
pip install pre-commit
pre-commit install  # this is global
```
then mess with Schutzfile and try to commit locally

schuellerf:
btw. we can also integrate this check in github like so:
https://github.com/osbuild/.github/blob/main/.github/workflows/pre-commit.yml#L20
(preferably with an up to date version :-D )

thozza (approved):
The change looks OK. I'm wondering if it would make sense to run tests, which are performed by the pre-commit config but missing in GH actions, as part of PRs? 

Mostly to prevent introducing new issues and makinbg life of anyone using the pre-commit hooks harder.

